PR #60, opened 2019-02-24
WIP: Make sure the browser does not show the access-key in the logs

NicolasDorier:
Attempt to fix https://github.com/ShahanaFarooqui/RTL/issues/58

### DIFF
--- a/controllers/authenticate.js
+++ b/controllers/authenticate.js
@@ -6,10 +6,25 @@ var upperCase = require('upper-case');
 var atob = require('atob');
 var logger = require('./logger');
 
-exports.authenticateUser = (req, res, next) => {
-  password = atob(req.body.password);
+exports.authenticateUserWithCookie = (req, res, next) => {
   if(+common.rtl_sso) {
-    if (common.cookie === password) {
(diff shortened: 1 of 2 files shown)


NicolasDorier:
NAAAAH wait I was testing it

ShahanaFarooqui:
yup, because i pull and tested it on my local and that failed :(

NicolasDorier:
btw, funny story: Actually in .8, if you give you `access_key` to a page which does not exist, it get properly processed and nothing appear in the browser!!

ShahanaFarooqui:
i was just confused if i should publish a release or not? :)

NicolasDorier:
I am testing, revert it :p

ShahanaFarooqui:
sure

NicolasDorier:
This is fun, actually even in .8, I can log in without trace. I just need to post to:
https://example.com/rtl/LOLOLOL?access_key=mykey

NicolasDorier:
The bug has unexpectedly the expected behavior.

ShahanaFarooqui:
For non-existent routes, node server will simply assume that the route exists within Angular. So for all non existent API request, we serve pre build angular with index.html. But, the access_key has to be correct :P

NicolasDorier:
reopened https://github.com/ShahanaFarooqui/RTL/pull/61

PR #61, opened 2019-02-24
Make sure the browser does not show the access-key in the history

NicolasDorier:
Reopening it!

### DIFF
--- a/controllers/authenticate.js
+++ b/controllers/authenticate.js
@@ -6,10 +6,30 @@ var upperCase = require('upper-case');
 var atob = require('atob');
 var logger = require('./logger');
 
-exports.authenticateUser = (req, res, next) => {
-  password = atob(req.body.password);
+exports.authenticateUserWithCookie = (req, res, next) => {
   if(+common.rtl_sso) {
-    if (common.cookie === password) {
+    res.cookie('access-key', req.query['access-key'], { signed: true, httpOnly: true, sameSite: true, secure: true });
+    res.set(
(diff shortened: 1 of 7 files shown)


ShahanaFarooqui:
Should i merge and publish it now?

ShahanaFarooqui:
Tested on my local, not working. Going offline for the night.

NicolasDorier:
almost there :)

NicolasDorier:
@ShahanaFarooqui where is the source code of the front end?

There is one thing missing: You need to include the `access-key` cookie in you request to `api/authenticate`

controllers/authenticate.js
```diff
@@ -6,10 +6,30 @@ var upperCase = require('upper-case');
...
 exports.authenticateUser = (req, res, next) => {
-  password = atob(req.body.password);
   if(+common.rtl_sso) {
-    if (common.cookie === password) {
+    const access_key = req.cookies['access-key'];
+    res.clearCookie("access-key");
+    if (common.cookie === access_key) {
       const token = jwt.sign(
         { user: 'Custom_User', lndConfigPath: common.lnd_config_path, macaroonPath: common.macaroon_path },
         'default_secret_key'
```
NicolasDorier: maybe we can use `common.cookieParserSecret` instead?

NicolasDorier:
also, the cookie is only for HTTPS!

NicolasDorier:
trying it

PR #62, opened 2019-02-24
Remove cookie signature

NicolasDorier:
Somehow, this does not work. But we don't really care about it actually.


ShahanaFarooqui:
@NicolasDorier Published with 0.2.10

PR #98, opened 2019-03-30
Extend the SSL access guide

openoms:
Add links to install Nginx and certbot.
Sample config for nginx.conf.

Successfully tested on a RaspiBlitz, but kept it generic.


openoms:
I have also made a script to set up the SSh access for RTL on the RaspiBlitz, but I think it is in a better place in that project`s menu or readme. 
https://github.com/openoms/bitcoin-tutorials/blob/master/nginx/README.md

saubyk:
Thanks @openoms for this.

Now get ready for support tickets! ;-)

PR #117, opened 2019-04-09
Add instructions to load ngx_stream_module.so in nginx.conf

goodc0re:


ShahanaFarooqui:
Thanks for updating the document.

PR #121, opened 2019-04-15
Update LND API Coverage

TomasHo:
Based on api.lightning.community


ShahanaFarooqui:
@TomasHo thanks for the PR.

PR #163, opened 2019-07-01
rename bitcoind.conf to default bitcoin.conf

seth586:
To avoid confusion, 

The default bitcoin configuration name is bitcoin.conf per https://github.com/bitcoin/bitcoin/blob/master/doc/bitcoin-conf.md


saubyk:
Thanks @seth586 :-)

PR #209, opened 2019-12-03
feat: add tooltip to publickey div

escapedcat:
make it easier for people who are not sure if this is the public key


ShahanaFarooqui:
Thanks.

PR #214, opened 2019-12-29
Update c-lightning instructions with service file

k3tan172:
Add service file instructions to c-lightning setup documentation

### DIFF
--- a/docs/C-Lightning-setup.md
+++ b/docs/C-Lightning-setup.md
@@ -8,6 +8,7 @@
 * [Installation](#install)
 * [Prep for execution](#prep)
 * [Start the server and access the app](#start)
+* [Run RTL as a service](#service)
 
 ### <a name="intro"></a>Introduction
 RTL can now be run to manage a c-lightning node via the UI.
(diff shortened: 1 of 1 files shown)


docs/C-Lightning-setup.md
```diff
@@ -101,5 +102,35 @@ If the server started successfully, you should get the below output on the conso
...
 Open your browser at the following address: http://localhost:3000 to access the RTL app.
 
+### <a name="service"></a>Running c-lightning RTL as a service (RPi or Linux Platform)
+In case you are running a headless Rpi or a Linux node, you can configure RTL as a service.
+
+* Create rtl.service systemd unit in and with the following content. Save and exit.
+```
+[Unit]
+Description=RideTheLightning
+Requires=c-lightning-REST.service
```
saubyk: This service dependency is applicable, only if c-lightning-REST is being run separately as a service.

If c-lightning-REST is being run as a plugin, than this dependency is not required.

k3tan172: Yes, that is correct. I might look into just setting this up as a c-lightning plugin than have it running as a service.

PR #304, opened 2020-04-09
Added hashbang

Kixunil:
This makes it easier to run the service just by running ./rtl.js

### DIFF
--- a/rtl.js
+++ b/rtl.js
@@ -1,4 +1,4 @@
-#!/usr/bin/node
+#!/usr/bin/env node
 
 const app = require("./app");
 const common = require("./common");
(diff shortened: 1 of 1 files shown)


Kixunil:
OMG, this is madness, there are conflicting standards :man_facepalming: 

It seems like `env` hack is preferable for people who like to use the newest versions of `node`, that might be installed outside packaging system, absolute path is preferred by packaging system, as it's more reliable (user can't break it by installing a different version somewhere else).

So the question is which group of people do you prefer to satisfy?

FYI, I'm doing packaging for Debian, but have no problem patching it with sed if you decide to honor the `env` style. It's still more robust than blindly inserting a line. :)

PR #359, opened 2020-06-10
feat: add channels visibility to channel tables

JosephGoulden:
It would be useful to see the channels visibility - public/private. I've added it as another column but happy to change it.

visibility


JosephGoulden:
Thanks @ShahanaFarooqui. I've now changed it so that the eye is next to the status as suggested. And it will only be displayed on the open channels tab.

ShahanaFarooqui:
Thanks, will review and merge the PR next week.

ShahanaFarooqui:
@JosephGoulden Thanks for the PR. We merged and extended this PR for ECL & CLT at [Release-0.8.4 branch ](https://github.com/Ride-The-Lightning/RTL/commit/928dba8ff2ee500de02bf8d406fd820e6e47e9df).

PR #361, opened 2020-06-11
WIP: Add Boltz as additional swap provider

dejankocev:
WIP: This PR adds [Boltz](https://docs.boltz.exchange/en/latest/) as an additional swap provider alongside [Loop](https://lightning.engineering/loop/). From UI perspective, **Loop in** (submarine swaps) and **Loop out** (reverse submarine swaps) have been renamed as **Deposit** and **Withdrawal** respectively, for sake of a more generalized terminology and paving the way for integration of additional swap providers. We welcome any sort of feedback including the ones associated with nomenclature. Boltz differs from Loop in that it doesn't require the user to set up an external daemon, which allowed us to integrate the Boltz API in the RTL directly. We've taken the utmost care to keep the end-user experience the same as that of swapping via Loop. Our decision to adhere to the same UI flow stems from the need to eradicate any user confusion that might come from otherwise different swap flow with different swap providers. We hope to have this PR reviewed, iterated on, and finally get merged into the master branch. 

Furthermore, we put out a [demo](https://twitter.com/Boltzhq/status/1262110466509115392?s=20) of a self-custodial so-called "channel creation swap" which we plan to offer as an API endpoint soon. We propose that we work on it next to have it added to the RTL. A channel-creation swap integrated into RTL will simplify node management for LN node operators as getting started with LN would be as simple as sending on-chain coins to Boltz and getting a balanced channel back. We'd love to hear your thoughts on it. 

TODOs for this WIP PR: 

- [ ] Edit after feedback

 TODOs for next proposed PR: 

- [ ] ACK for the UI for channel creation swap
- [ ] Integrating channel creation swap


ShahanaFarooqui:
Thanks for the PR. We are currently working on Eclair implementation and it should be finished by the end of this week. After that, we will try to merge it ASAP.

armurbalda:
Did you guys have a chance to look at this yet? @ShahanaFarooqui

ShahanaFarooqui:
Saubyk has already started testing it. I will start today.

ShahanaFarooqui:
Hi, I went through the PR and below are some initial observations regarding coding and architecture.
We will provide more feedback wrt design adjustments which needs to be made in addition to the technical changes.

1) There were a lot of new additions and code cleanups due to Eclair implementation. I merged those changes into the PR. Please check and verify that none of your code is broken/missing.

2) We are conservative about adding new package dependencies in the project and new additions are made based on absolute necessacity.
	2.1) Swap services are optional, the dependencies for Boltz should be installed only when the users opts in for it, the users should not be installing these additional dependencies by default. We would prefer to add instructions in the documentation for Boltz user to install these dependencies locally, when needed. 
	2.2) bitcoinjs-lib: bitcoinjs-lib has indirect dependency on node-gyp which uses 'Python2'. This dependency has proved to be a headache for most of the node implementations like RaspiBlitz and NODL and it might raise a lot of issues. We highly recommend to use a library which does not use node-gyp or convince the node implementations to install 'Python 2'. Without either of these two, the solution will not work.
	2.3) Where are 'process' & 'stream' libraries being used? If there is no use of these libraries you should remove them.
	2.4) Do you really want to install 'buffer' explicitly? Angular should already be able to use Nodejs's buffer after "node/types" & polyfills are in place.

3) Any swap service is optional and has to be enabled by the user explicitly. With current code, user will not be able to opt-out from Boltz. The design has to be adjusted to make it an optional service, which the user needs to choose.

4) We need to keep Loop and Boltz logic, pages and menu separate.

5) BOLTZ_API_URL must be constructed based on testnet/mainnet. Currently, it is hard coded with testnet url.

6) BOLTZ_LND_NODE must be configurable. Currently, it is hard coded with mainnet address.

7) boltzSwapsPath and other configurable variables should be available for environment variable configuration also. Environment variables get higher precedence over config file. 

8) When I tested the PR, Boltz withdrawal failed with error : 
	ERROR TypeError: Cannot read property 'call' of undefined
    at Hash.CipherBase (index.js:7)
    at new Hash (browser.js:9)
    at createHash (browser.js:29)
    at Object.sha256 (crypto.js:23)
    at BoltzService.push../src/app/shared/services/boltz.service.ts.BoltzService.onSwap (boltz.service.ts:52)

9) There were/are some minor coding related issues, like:
	9.1) ngrx action 'SEND_COIN' was same as 'SetChannelTransactions', So I removed the new action and used the existing one.
	9.2) LightningLab Swaps are not showing in the table, only boltz swaps are visible.
	9.3) Date/time format is not standardized.
	9.4) Infographics on the modal is loop specific. So, either Boltz should have its own infographics, or the button to show infographics must be hidden.

10) Is there any Medium post or documentation which explains the complete swap logic. Please provide any documentation references detailing the functional description of Boltz swaps. This is required for us to understand the service better, as well include in the app for references, so that user can read and understand the details of the service.

@saubyk  is currently working to understand the Boltz flow and to design the architecture accordingly. He will provide his observations later.

michael1011:
Hi @ShahanaFarooqui . Thanks for the thorough review, highly appreciated! It was exactly our expectation that we can take this initial version and modify it into something that would meet your standards so it can be part of RTL.

1) We'll check the flow and make sure nothing is broken

2.1) Noted! We're already looking into making the adjustments so this can be a opt in functionality
2.2) Indeed a headache, we are looking into it.
2.3, 2.4) Think these issues will be solved by updating the boltz-core version which will be stripped of these libraries as they are not necessary for the functionality we support in RTL

3) // this feels the same as 2.1

4) Noted. We'll separate the logic into it's own menu and it won't be mixed. Product-wise we see the benefit of the user having a "Select a provider" UX though. Wdyt?

5) We're all clear on this one and we'll make the change

6) All clear on this one and we'll make the change here as well

7) So if we understand correctly the priority goes like this: env variable -> config file -> default value?

8) We will look into it and resolve the error

9) We'll resolve these

10) 
- docs: https://docs.boltz.exchange/
- How boltz submarine swap works: https://medium.com/boltzhq/submarine-swaps-c509ce0fb1db
- How hold reverse submarine swap works: https://medium.com/boltzhq/brewing-boltz-hold-reverse-swaps-fea0fadbf041
- or directly get in touch with us: https://discord.gg/QBvZGcW

saubyk:
Hi Folks,

Getting excited about this PR. I believe this is a good complementary service, which should be provided to routing node operators. Noting down my initial product recommendations. We should seek design guidance from @diogorsergio also, as we are making substantial changes to the product organization with this PR.

Nomenclature of service:
`Swap Out` and `Swap In` are better labels for Boltz swap actions instead of Withdrawal or Deposit. Since this is essentially a swap between off chain and on chain funds, not a unidirectional transaction.

Service selection by the user:
User should enable Boltz service by a config option in the `RTL-Config.json` file (In the long term we should enable this config option in the UI settings). For now, guidance for this config setting will be provided in the documentation. We can explore developing this feature in the settings as well, enabling it for both Loop and Boltz.

Menu placements in the app:
- When the swap is initiated from the dashboard. Service selection options should be presented only if the user has opted in for multiple services.
- Side navigation should have a menu, ‘Optional Services’, at the same level as ‘Lightning’, which should be enabled when the users opts into either Loop or Boltz or both.
- Boltz and Loop will be different sub menus under ‘Optional Services’.
- Actions menu on the channels grid should also have ‘Swap Out’ menu, which opens up the same modal as the one from the merchant dashboard for channel specific swap out.

Action Flow:
Mapping a high-level user journey below.
1. User enables Boltz via config option/config settings
2. Swap menus are enabled in the application dashboard and side navigation menu
3. Swap Out option can be accessed from the Merchant Dashboard (outbound liquidity) or Channels Action menu or Swap service menu.
4. Swap In option can be accessed from the Swap service page
5. Status updates of swap transactions should be updated on the grid once the transaction has been placed.

Other Lightning Implementations:
Since Boltz is not a LND specific service, you should make Boltz a common component in RTL. This would enable you to offer this service to C-Lightning and Eclair users along with LND, expanding your potential user base. @ShahanaFarooqui  can provide guidance on how to make it a common service/component in RTL architecture.

Infographics:
Consider creating your custom infographics to explain the swap process to the user. Swaps are complicated and having an educational aid can be helpful for adoption and usage.

Will be providing more inputs as we make progress here.

saubyk:
Please refer to this issue for design guidelines that we are working on...
https://github.com/Ride-The-Lightning/RTL-Design/issues/54

armurbalda:
This is ready to be reviewed once more @saubyk @ShahanaFarooqui @diogorsergio

saubyk:
> This is ready to be reviewed once more @saubyk @ShahanaFarooqui @diogorsergio

Hey guys, will start the review of this PR after the v0.9.1 release.

Looking forward to testing it out, soon.

ShahanaFarooqui:
Future Boltz PR updates will be maintained with PR https://github.com/Ride-The-Lightning/RTL/pull/540

ShahanaFarooqui:
Closing it with PR #590

PR #401, opened 2020-08-05
rename 'Receive from Peer' to 'Send to channel'

paboset:
As a new user & LN rookie,
'Receive from Peer' was confusing.
I thought it was trying to have me receive on the channel I clicked 'Circular rebalance'
but the balance limit on the `Amount` field implied the opposite (of course).
It caused a few brain cycles to spin.
I even did the opposite of what I wanted the first time.

Does 'Send to channel' make the choice more clear
that I'm sending From the channel I clicked on
To the channel I'm picking from,
to rebalance?

image


saubyk:
Hi @paboset 
I agree this can be a bit confusing, but I still believe `Receive from Peer` label does a better job of explaining what is happening.
Circular Rebalance is you paying yourself, through a particular route.
1. You chose the channel you send the payment out from
2. You select the node you'll receive the payment back from (note that you cannot specify a channel to receive from, but a peer)

Since you cannot specify the channel you are receiving on, it should not be labeled as `Send to Channel`. E.g. if you have multiple channels with the same node, you will not be able to control which channel you get the payment back on.

All this being said, we can add more explanation on the modal window to help grasp the concept.

ShahanaFarooqui:
@paboset Thanks for raising the issue and PR. Closing it via PR #449.

PR #424, opened 2020-08-17
[CLT] group payments for v0.9.x

### LINKED ISSUE #419: [CLT] v0.9.0 api update lists multiple entries for a payment hash
`listsendpays` functionality has been updated.  Previously it would list the current state as either `success` or `failed` for a given payment hash.  With the introduction of mpp, it now list each attempt, fail or success, using the `partid` field with a new id for each attempt.  This could potentially list many failed attempts.  Also, if a large payment is sent, it could be split into multiple paths, and they each now show with `partid` 1-nn

To reproduce, the c-lighning node needs to be upgraded, and you need to send payments that do not succeed on the first attempt, or succeed via multiple paths

Screenshot from 2020-08-13 19-26-14
*this was a single attempt to pay using `pay` command*

RTL 0.8.4-beta

`listpays` provides the grouping but not all fields, most notably, `created_at`

the grouping for `listpays` is shown in [this loop](https://github.com/ElementsProject/lightning/blob/3d902501e5593b24f26c4bab0ff2d7ca652afc16/plugins/pay.c#L1798)

I will follow up with a pull request in the coming days

rsbondi:
This fixes #419.   This should be backward compatible for previous c-lightning versions, it will just group each as a single payment.  I am using the most recent time value of all attempts.

Note that there is an issue in the release where the `destination` is not being recorded with payments, it has been fixed [here](https://github.com/ElementsProject/lightning/pull/3888).  When building c-lightning in the current repo state, the `destination` appears as usual in RTL

### DIFF
--- a/controllers/c-lightning/payments.js
+++ b/controllers/c-lightning/payments.js
@@ -3,6 +3,35 @@ var common = require('../../common');
 var logger = require('../logger');
 var options = {};
 
+function _groupPayments(payments) {
+  return Object.values(payments.reduce((result, current) => {
+    const multi = !!current.partid;
+
+    result[current.payment_hash] = result[current.payment_hash] || {
+      id: current.id,
+      payment_hash: current.payment_hash,
+      label: '',
+      created_at: current.created_at
+    };
+    const payment = result[current.payment_hash];
+    if (!multi) payment.id = current.id;
+
+    if (current.payment_preimage) payment.payment_preimage = current.payment_preimage;
+    if (current.bolt11) payment.bolt11 = current.bolt11;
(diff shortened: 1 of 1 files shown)


rsbondi:
@saubyk The first two bullets were deliberate, this is mimicking `listpays`, but I am checking with the c-lightning telegram group, I don't wee an obvious way to group, but hopefully they can help.

I am able to get the dialog in my setups for all payment types, mpp, keysend and legacy.  Maybe there is certain data on some transactions that it does not handle?  Or something else, anything you can tell me to help me duplicate?

Issue #410, opened 2020-08-11
Can't select priority on channel close


sendbitcoin:
When I want to close a channel i can't click on priority to change it. no dropdown appears and i cant select an other option then default

RTL version 0.8.4
Firefox version 79.0
Chrome version 84.0.4147.125

![2020-08-11 19_02_30-192 168 1 195 - Remotedesktopverbindung](https://user-images.githubusercontent.com/60074674/89926481-43e13000-dc05-11ea-81f9-e16fec3816ba.png)

ShahanaFarooqui:
@sendbitcoin We disabled priority options if the channel is inactive because LND doesn't allow priority values when the channel is inactive. Please let us know if it is disabled for active channel as well.

ShahanaFarooqui:
Related with #365

sendbitcoin:
Yeah i just figured that out closing the channel with lnd.

maybe don't display the dropdown on inactive channels and instead display the message that lnd gives:

force closing a channel uses a pre-defined fee


or even better would be if you could display exactly what that pre-defined fee will be if I force close a channel, but not sure if lnd gives that info

ShahanaFarooqui:
@sendbitcoin 
`or even better would be if you could display exactly what that pre-defined fee will be if I force close a channel, but not sure if lnd gives that info`
You are correct, LND neither provides the information not let us set the value for inactive channels.

`maybe don't display the dropdown on inactive channels and instead display the message that lnd gives?
force closing a channel uses a pre-defined fee`
- Removing the dropdown will be against our current UX. 
- Showing message is a great suggestion. Assigning the issue to Diogo for further advise.

sendbitcoin:
>Removing the dropdown will be against our current UX.

as a user when i see a dropdown that i can't click on and can't interact with at all I immediatly think it's a bug/error

maybe displaying the message will solve the confusion, but I still think if you can't interact with it, it shouldnt be displayed at all (or at least be greyed out or something to indicate it's supposed to be not-clickable)

ShahanaFarooqui:
Done with PR #444.

Issue #393, opened 2020-08-02
pending_htlcs is not handled when closing channel


zorbamorba:
**Describe the bug**
If lnd node  is routing a transaction and user try to close the channel RTL show unknown error try later .  

**To Reproduce**
cleate a pending_htlcs
try to close the channel which is routing the transaction .

**Your environment**
* Version of `RTL` 0.8.2
* Version of `lnd` 0.10.3
* Version of `bitcoind` 0.20 

**Additional context**
pending_htlcs information should be in channel info

saubyk:
@ShahanaFarooqui this can be handled in the following manner:
1. Upon click of the `Close Channel` action, fetch the channel information with the `listchannels` api.
2. If the channel has pending HTLCs in the response json, throw an error to the user
3. The error should state: `Uncleared HTLCs currently pending within the channel. Please wait for the HTLCs to clear before attempting channel closure`

ShahanaFarooqui:
Done with PR #444.

PR #444, opened 2020-08-23
Channel closing with pending htlc & inactive without priority #393 & #410

ShahanaFarooqui:
Channel closing with pending htlc & inactive without priority #393 & #410


saubyk (commented):
Tested again. Looks ok.

Issue #453, opened 2020-08-26
Bug: Channel and peer counts truncate if 3-digits and large browser font


aantonop:
**Describe the bug**
For those of us who need reading glasses and set our browser text to large, and who also run nodes with > 100 channels and peers, the channel/peer counts in the little circles, are truncated to one digit and "..."

**Screenshots**
![image](https://user-images.githubusercontent.com/1181915/91353244-0067f380-e7b9-11ea-996e-acfd4a35381b.png)

**Your environment**
* Version of `RTL`: 0.8.2.beta
* Version of lnd 0.11
* Version of bitcoind 0.20
* Browser & browser version Firefox and Chrome
* Operating system Ubuntu 20.04 LTS

**Additional context**
Try not to "fix" this by making the font smaller - don't make me have to put on glasses to ride the lightning. Maybe make the circle larger?

aantonop:
Confirmed also in RTL v0.9.0

ShahanaFarooqui:
Closing the issue with PR #480. The fix will be released in v0.9.1.

![image](https://user-images.githubusercontent.com/23240721/93957097-89733b80-fd21-11ea-832c-a7cd08c68640.png)

PR #480, opened 2020-09-23
Increase badge width with number #453

ShahanaFarooqui:
Increase badge width with number #453

Now the badge width will increase with number.
image


saubyk (approved):
Tested. Looks good.

Issue #451, opened 2020-08-26
Provide feature to sweep all on-chain funds on C-Lightning


saubyk:
The api for on-chain withdrawal now supports ability to withdraw all the on-chain funds from the c-lightning node.
This can be added as another type of on-chain transaction, similar to what we have in LND.

ShahanaFarooqui:
Closing with PR #483

PR #483, opened 2020-09-23
CLT Sweep All #451

ShahanaFarooqui:
CLT Sweep All #451


saubyk (approved):
Tested. Looks good. Ok to merge.

Issue #454, opened 2020-08-27
Incorrect USD formatting when paying invoice


BTCBellyButton:
**Describe the bug**
The USD formatting when sending a payment is incorrect.  See Screenshot.
In USD it should read  $34.20951 (or just $34.21 if rounded). (dot as decimal point and comma for thousands)
As it is now it means 34 thousand 2 hundred and 9 dollars, and 51cents.

**To Reproduce**
Pay a 300,000 sat invoice. 

**Screenshots**
![image](https://user-images.githubusercontent.com/68573637/91382924-76d41800-e7f0-11ea-93c7-48afa2ec8f8c.png)

**Your environment**
* Version of `RTL`:  0.9.0
* Version of `c-lightning`:  v0.9.0-119-g7435d50
* Version of  `bitcoind`, : 0.20.1
* Browser & browser version: Dissenter Version 1.4.95 Chromium: 80.0.3987.122 (Official Build) (64-bit)
* Operating system (`uname -a` on *Nix): Windows 10 / Ubuntu 20.04.1 LTS

ShahanaFarooqui:
Closing with PR #484.

PR #484, opened 2020-09-23
CLT Invoice incorrect fiat conversion #454

ShahanaFarooqui:
CLT Invoice incorrect fiat conversion #454


saubyk (approved):
Tested. Issue is fixed. Approved to merge.

Issue #489, opened 2020-09-25
Pull the description from the getsentinfo to show in the payment view info


ShahanaFarooqui:
For the payments, the description information in the payment request is useful for the user to have a descriptive link to the payments made. This feature has been made available on C-Lightning, same can be done for ECL as well.

ShahanaFarooqui:
Closing it with PR #490.

PR #490, opened 2020-09-25
ECL Added Description in Payments Info #489

ShahanaFarooqui:
ECL Added Description in Payments Info #489


saubyk (approved):
Tested. looks good to merge.

Issue #472, opened 2020-09-20
loop integration broken


delistandecrypt:
Since loop [v0.9.0-beta](https://github.com/lightninglabs/loop/releases), it uses TLS for the API.
 When trying to access loop from RTL, an error is printed in the logs:
```
Sep 20 19:12:16 endor node[745067]: ERROR: Loop(316) =>
 List Swaps Error: {"name":"RequestError","message":"Error: self signed certificate","cause":
{"code":"DEPTH_ZERO_SELF_SIGNED_CERT"},"error":{"code":"DEPTH_ZERO_SELF_SIGNED_CERT"},"options":
{"url":"https://localhost:8081/v1/loop/swaps","simple":true,"resolveWithFullResponse":false,"transform2xxOnly":false}}

```

ShahanaFarooqui:
Closing with PR #491

PR #491, opened 2020-09-25
Loop upgraded to 9.0 with backward compatibility #472

ShahanaFarooqui:
Loop upgraded to 9.0 with backward compatibility #472


marssystems:
Just tested this and it gets stuck on the spinner  with the message "Getting Terms and Quotes".
This happens on both the Loop In and the Loop Out.

ShahanaFarooqui:
> Just tested this and it gets stuck on the spinner with the message "Getting Terms and Quotes".
> This happens on both the Loop In and the Loop Out.

@marssystems Thanks for testing the PR. You can now test it again.

saubyk (approved):
Tested. Approved for merge.

Issue #441, opened 2020-08-22
ECL Include amount and amount settled in the invoice list


ShahanaFarooqui:
The invoice list should show requested value of the invoice and amount settled as separate fields for Eclair implementation.

Blocked due to open issue with Eclair. ACINQ/eclair#1511

PR #492, opened 2020-09-25
ECL Added Settled Amount in Invoices list #441

ShahanaFarooqui:
ECL Added Settled Amount in Invoices list #441


saubyk (approved):
Tested. Looks good to merge.

Issue #420, opened 2020-08-14
[LND] List the UTXOs for the wallet


saubyk:
UTXOs from the LND wallet, can be listed under the `On-Chain` menu.
It should be a tab next to the `Transaction History` page, showing the UTXOs of the wallet.
Enable this feature for LND v11 and above as the v1 api is deprecated.
The API to be used for this: `/v2/wallet/utxos`

ShahanaFarooqui:
Done with PR #493

PR #493, opened 2020-09-25
List the UTXOs for the wallet #420

ShahanaFarooqui:
List the UTXOs for the wallet #420


saubyk (commented):
UTXOs API may have an issue. Not getting any records for the UTXOs with the current API

saubyk (commented):
The grid for UTXOs should list TXID and Output.

saubyk (approved):
Tested. Approved to merge.

PR #502, opened 2020-10-01
Confirmation message updated for force close

ShahanaFarooqui:
Confirmation message updated for force close


saubyk (approved):
Tested. Approved to merge.

Issue #497, opened 2020-09-27
Path information missing for transaction (under view info)


sendbitcoin:
I upgraded to 0.9.0 and now the transaction page shows the number of hops again in the column... but if I click on view info I cannot see the actual peers that were involved in the transaction.

currently I have no way of figuring out what path a transaction took.

for failed transactions the hops are displayed correctly under view info. (and I can scroll down infinitely)

ShahanaFarooqui:
Closing with PR #504

PR #504, opened 2020-10-06
Path information missing for transaction #497

ShahanaFarooqui:
Path information missing for transaction #497


saubyk (approved):
Looks good. Can be merged. Thanks

Issue #383, opened 2020-07-16
Problem parsing eclair.conf


n1bor:
**Describe the bug**
The eclair config is HOCON format:
https://github.com/lightbend/config/blob/master/HOCON.md

so:

```
eclair.api.password=test1234
```
is the same as
```
eclair {
  api {
   password=test1234
  }
}
```
but you are not parsing it correctly if file is in the later format.

**To Reproduce**
Enter password in eclair.conf as per the second code sample and the config in RTL is confused.

You could just document this and people could rewrite their config so is in the a.b.c format.

ShahanaFarooqui:
@n1bor you can use ‘lnApiPassword’ key directly under Authentication rather than ‘configPath’. It is available for v0.8.2.

ShahanaFarooqui:
@n1bor Were you able to fix the issue with lnApiPassword?

ShahanaFarooqui:
Closing with PR #505.

PR #505, opened 2020-10-06
Problem parsing eclair.conf #383

ShahanaFarooqui:
Problem parsing eclair.conf #383


saubyk (changes requested):
Please add the parser as a dependency.

Issue #517, opened 2020-10-20
Unable to Decode Payment


ShahanaFarooqui:
The application is unable to decode the payment, due to recent RTL server's endpoints updates for the API.

ShahanaFarooqui:
Closing with #518

PR #518, opened 2020-10-20
ECL Decode Payment #517

ShahanaFarooqui:
ECL Decode Payment #517


saubyk (approved):
Tested and approved for merge.

Issue #519, opened 2020-10-21
Asking for my RTL password, but in single sign on, there is no password


NicolasDorier:
I can't sweep my funds.

![image](https://user-images.githubusercontent.com/3020646/96722803-bda54f00-13e8-11eb-96ad-b23bbf5dd008.png)

saubyk:
Hi @NicolasDorier fix is now available with the release v0.9.2

PR #520, opened 2020-10-21
Sweep All asking for password from SSO users #519

ShahanaFarooqui:
Sweep All asking for password from SSO users #519


saubyk (approved):
merge

Issue #529, opened 2020-10-28
Add support to bump fee for channel open on LND


saubyk:
Mempool is getting pretty busy these days and channel opening transactions can get stuck if unsuspecting users low ball the fee rate.
Explore the fee bump API to alter the fee rate and raise the priority of the channel opening transactions.
GRPC: https://api.lightning.community/?javascript#bumpfee
REST: /v2/wallet/bumpfee

Reference:
https://lightningwiki.net/index.php/Bumping_fee_for_lightning_channel_open

Feature description:
- The fee bump should be enabled for the `Pending Open` section on the Pending tab of the Channels page via the Actions menu.
- Upon click of the 'Bump Fee' button, user will be presented with a modal to take further action.
- The following text should be presented as information to the user
"Bumping fee on pending open channels is an advanced feature, attempt it only if you are familiar with the functionality of Bitcoin transactions.
Before attempting fee bump ensure the following:
1. Use a Bitcoin block explorer to ensure that channel opening transaction is not confirmed.
2. The channel opening transaction must have a sizable change output, which can be spent further.
3. Confirm the index value of the change output via a block explorer.
4. Upon successful fee bump, use the block explorer to verify transaction inclusion on the block"
- Present the user with the modal to enter the `Index value of change output`, `Fee rate`/`Target Blocks`
- Validate the inputs and invoke the API to bump fee.
- Upon successful execution, inform user to verify transactions on the block explorer.

viaj3ro:
could this be added for eclair as well? 

Is it technically possible, to also cancel stuck funding TX by either sending back to self or even better CPFPing them for a different channel funding tx?

saubyk:
> could this be added for eclair as well?
> 
> Is it technically possible, to also cancel stuck funding TX by either sending back to self or even better CPFPing them for a different channel funding tx?

Don't think Eclair has an API to bump the fee. If they do, will be happy to add this feature for Eclair.
Do note that bumping of fee on pending channel open is conditional upon availability of change output or an ability to do coin selection for channel open.
Let me describe the two scenarios:
Scenario # 1: Channel Open has a change output
- Spend the change output to self with a higher fee rate

Scenario # 2: Channel Open has specific UTXOs
- Reopen the channel with the same UTXOs but at a higher fee-rate

viaj3ro:
Does eclair has such an API @t-bast ?
If not, are you willing to ad it?

ShahanaFarooqui:
Closing with PR #531

[...]

PR #531, opened 2020-10-29
Open Channel Bump Fee #529

ShahanaFarooqui:
Open Channel Bump Fee #529


saubyk (changes requested):
- Open Channel, advanced option failing to load the fee options in FireFox
- Index value of 0 is not working for fee bump
- Correct "Bump fee for channel" to "Bump fee for channel point" on the modal
- Correct "Output Index" label to "Index for change output"
- Transaction type selection is not rendering and requires browser reloads on FireFox

saubyk (approved):
Approved for merge

Issue #538, opened 2020-11-15
Add label to Loop APIs


ShahanaFarooqui:
Add 'RTL' label in Loop APIs to inform server that the loop is initiated from RTL.

ShahanaFarooqui:
Closing with #543.

Issue #515, opened 2020-10-14
Loop v0.10.0 - beta not working with RTL v0.9.1-beta


marssystems:
**Describe the bug**

I'm getting an Error 500 - Client sent an HTTP request to an HTTPS server - when running the latest version of Loop (0.10.0) with version 0.9.1 of RTL.

ShahanaFarooqui:
What is your swapServerUrl value in RTL-config.json file? Loop server runs on https on version 0.10.0. Please update the url from http to https, restart RTL server and test again.

marssystems:
Yep! - changing this to https solved the problem.

Now I am getting - expected 1 macaroon, got 0 . Searching for loop.macaroon shows it residing in /home/loop/.loop/mainnet/ but for some reason Loop is unable to find it. It could be some quirk of Razpiblitz.

saubyk:
@ShahanaFarooqui Loop v0.10.0 implemented macaroons.

@marssystems you'll need to wait for the next release, so that we add macaroon support for Loop.
For now, please use Loop v0.9.0 which has been tested and supported on RTL v0.9.1

Kixunil:
Any ETA on this? I'd like to use loop with RTL but so far I avoided it because the absence of authentication made it unsecure. Solving this would unblock me.

ShahanaFarooqui:
Closing with #543.

openoms:
> What is your swapServerUrl value in RTL-config.json file? Loop server runs on https on version 0.10.0. Please update the url from http to https, restart RTL server and test again.

@ShahanaFarooqui thsi needs to be updated in the sample-config file: https://github.com/Ride-The-Lightning/RTL/blob/6e036d838255124d83a33d55e52b9d8e0aef933a/sample-RTL-Config.json#L26

PR #543, opened 2020-11-18
Added macaroon authentication for Loop

ShahanaFarooqui:
- Added macaroon authentication for Loop #515
- Added RTL Label for Loop In & Loop Out #538


saubyk (approved):
Approved for merge

PR #562, opened 2020-12-10
CLT & ECL Reports

ShahanaFarooqui:
CLT & ECL Reports


saubyk (changes requested):
Please take care of the below items before merging:
- Transactions report should not show "Received 0 Sats/0 Invoices" if the received sats is zero. Vice versa for the Payments.
- The charts are not rendering on chrome browser

Issue #542, opened 2020-11-18
Mitigate brute force attacks with fail2ban


bota87:
I found that RTL authentication is vulnerable to brute force attacks so I tried to mitigate this by using fail2ban but actualy this is not possible because the ip address of the host that made the request is not logged with the authentication error.
`nov 18 23:26:59 server RTL[18877]: ERROR: Authenticate(36) => Invalid Password!`
If you add the ip address in the log message than I can create a filter for fail2ban and ban that address after X failed attempt

ShahanaFarooqui:
@bota87 Thanks for the issue. We are logging failed IP addresses now. Additionally, We implemented the logic to lock the application for 30 minutes after 5 unsuccessful attempts.

Please note that if the RTL is served through proxy server, then it has to set 'X-Forwarded-For' header also. NGINX example:
`proxy_set_header X-Forwarded-For $remote_addr;`

Closing with PR #566.

PR #566, opened 2020-12-15
App lock after 5 attempts #542 & DatePicker default adapter #532

ShahanaFarooqui:
App lock after 5 attempts #542 & DatePicker default adapter #532


saubyk (approved):
Tested. Can be merged.

PR #567, opened 2020-12-16
fix loop auth details in the sample config

openoms:
as discussed in: https://github.com/Ride-The-Lightning/RTL/issues/515#issuecomment-745980140

and changed in: https://github.com/Ride-The-Lightning/RTL/pull/543


ShahanaFarooqui:
@openoms Thanks for the PR. However, the documentation, including sample-RTL-Config' was updated with the release. Closing the PR now.

PR #609, opened 2021-02-23
Remove node-sass from README

AaronDewes:
Since #377, node-sass isn't a dependency anymore


ShahanaFarooqui (approved):
Thanks for the update and PR.

PR #612, opened 2021-02-27
Fixed security edge case

### LINKED ISSUE #610: Security risk: RTL can be accessed with empty SSO key when the space runs out
**Describe the bug**

If the partition used for storing the SSO cookie runs out of space, the SSO token is set to empty leading to enabling trivial access by anyone. LN node may still be working if the partition is different (as was in my case), so stealing is possible. Non-private disclosure was selected because this can only happen in edge case scenarios and should be very hard to cause intentionally.

The bug occurs because the ENOSPC error is only generated by the OS when writing to a file, not when it's created.  The best practice to avoid these kinds of errors and other catastrophic data corruption bugs is to write a temporary file in the same directory, sync it (fsync system call) and if all that succeeded move it over the old location. Move is atomic, so the data can't get corrupted. In case of RTL, unlinking the cookie file before attempting to do anything else sounds like a good way to prevent reuse of the cookie.

Additionally, I recommend to **not** read the cookie from the file at all, just write it. Remember and use the copy that is already in memory that was used for writing. This way no weird kind of OS failure can ever cause the cookie to be predictable by the attacker. Writing the code this way would've prevented this attack.

Feel free to take inspiration from a similar [code I wrote for `btc-rpc-explorer`](https://github.com/janoside/btc-rpc-explorer/commit/28e40387d2b32b652461ac1ad34d1799ba77b5b9) that was written defensively, using both mentioned techniques. Although now I see it doesn't contain explicit fsync call - not a big deal there since it actually doesn't need to be physically stored on disk (Cache in RAM is enough, next boot will invalidate it anyway.) Note that token is always refreshed on start - this is intentional to avoid surprises like this one.

The code mentioned above happens to also have more optimal token length and representation.

**To Reproduce**
1. Install and configure RTL to use SSO, pointing to a different partition than LN node
2. Fill up the prtition
3. Attempt to login - it will fail
4. Attempt to login with empty key - it will succeed

**Your environment**
* Version of `RTL`: 0.9.0 - I already checked that the cookie handling code is broken in master
The rest is irrelevant as the code is OS-independent.

**Additional context**
The cookie handling is implemented in:

https://github.com/Ride-The-Lightning/RTL/blob/f817ae39bc71594f4e4a7a6be8cf68c74ac2362c/connect.js#L301-L333

Kixunil:
This fixes the edge case that'd lead to a vulnerability when the storage
fills up.

Closes #610

### DIFF
--- a/connect.js
+++ b/connect.js
@@ -272,7 +272,7 @@ connect.setSSOParams = (config) => {
     if (!common.rtl_cookie_path || common.rtl_cookie_path.trim() === '') {
       errMsg = 'Please set rtlCookiePath value for single sign on option!';
     } else {
-      connect.readCookie(common.rtl_cookie_path);
+      connect.refreshCookie(common.rtl_cookie_path);
     }
   }
 };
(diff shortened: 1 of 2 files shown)


ShahanaFarooqui:
@Kixunil, Thanks for the PR. However, as explained in the issue #610, we are closing it with #646.

PR #630, opened 2021-03-15
Updated instructions for TOR and SSL (Debian based distros)

PsySc0rpi0n:
I have upated the instructions for TOR and SSL configuration instructions. The existent ones were kind of outdated. I only updated for Linux (Debian based distros) and Android (for TOR).

Any improvements are welcome. I'm no expert and I might have made mistakes. This was my first contribution, if it can be called as such, to a "real" repository!

Hope it is useful!
Thanks
Psy

### DIFF
--- a/docs/RTL_SSL_setup.md
+++ b/docs/RTL_SSL_setup.md
@@ -39,3 +39,5 @@ Sample configuration to be inserted in the nginx.conf (adjust the path and filen
     }
 
 Restart Nginx with the new configuration and connect to RTL over https on the port 3002.
+On Debian based distros:
+    $> sudo systemctl restart nginx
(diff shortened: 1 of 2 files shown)


bitbandi:
if you use nginx for ssl gateway only, it's better is you use ssl in nodejs directly:
https://stackoverflow.com/questions/5998694/how-to-create-an-https-server-in-node-js

PsySc0rpi0n:
> if you use nginx for ssl gateway only, it's better is you use ssl in nodejs directly: https://stackoverflow.com/questions/5998694/how-to-create-an-https-server-in-node-js

I have this setup with nginx now. So, I'm not even going to think about changing. heheh. But thanks for suggesting. I just tried to update instructions for what was already here!

bitbandi:
if you have already an nginx server with many vhost, than it's a good choice to use as reverse proxy. But if you run nginx for rtl only, it's "useless", yet another thing that can go wrong (murphy's law :D )

mdconaway:
Hey @bitbandi
I just added PR #1026
This should add pretty much all of the possible SSL features.

Issue #636, opened 2021-03-29
unencrypted password


plickami:
Under LND config in the settings is visible unencrypted password "tor.password=*******" used for logging in to the RTL (raspiblitz password B)!

To Reproduce
1. log in to RTL
2. Go to: settings -> LND config -> scroll down
3. See "tor.password" unencrypted password used for RTL login

environment:
RaspiBlitz v1.6.3 
Bitcoin Fullnode + Lightning Network + Tor
RTL version 0.10.0 beta
Browser Edge and/or firefox

ShahanaFarooqui:
Closing with PR #647.

PR #647, opened 2021-04-11
Mask config file passwords #636

ShahanaFarooqui:
Mask config file passwords #636

### DIFF
--- a/controllers/shared/RTLConf.js
+++ b/controllers/shared/RTLConf.js
@@ -165,7 +165,7 @@ exports.getConfig = (req, res, next) => {
   logger.info({fileName: 'RTLConf', msg: 'Node Type: ' + req.params.nodeType + ', File Path: ' + confFile});
   fs.readFile(confFile, 'utf8', function(err, data) {
     if (err) {
-      logger.error({fileName: 'Conf', lineNum: 171, msg: 'Reading Conf Failed!'});
+      logger.error({fileName: 'Conf', lineNum: 168, msg: 'Reading Conf Failed!'});
       res.status(500).json({
         message: "Reading Config File Failed!",
         error: err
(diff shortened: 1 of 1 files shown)


controllers/shared/RTLConf.js
```diff
@@ -335,3 +304,22 @@ exports.updateServiceSettings = (req, res, next) => {
...
+  if (length !== 0) {
+    for (var i = 0; i < length; i++) {
+      if (typeof obj[keys[i]] === 'object') {
+        keys[keys[i]] = maskPasswords(obj[keys[i]]);
+      }
+      if (typeof keys[i] === 'string'
+        && (keys[i].toLowerCase().includes('password') || keys[i].toLowerCase().includes('multipass')
+          || keys[i].toLowerCase().includes('rpcpass') || keys[i].toLowerCase().includes('rpcpassword'))
+      ) {
+        obj[keys[i]] = obj[keys[i]].replace(/./g, '*');
```
Kixunil, line 320: Leaking password length is also not great but I guess whole issue was exaggerated as anyone with access to RTL could just steal sats directly, right?

ShahanaFarooqui: @Kixunil Thanks for the review. 

As you already stated, if someone is able to access upto this layer then the user has bigger issues to worry about than the length of the password. :)

Additionally, this `length` variable is NOT reading the length of the password but only the length of the keys. ie. the number of sub-keys available under that header/key. 

For example, the length will be 3 for below header (`'Bitcoin`) as it has three sub-keys.
[Bitcoin]
bitcoin.active=1
bitcoin.testnet=1
bitcoin.node=bitcoind

Kixunil: I meant leaking length by replacing each character with exactly one asterisk as opposed to replacing whole password with fixed-length string. (One can determine password length by counting the number of asterisks.)

ShahanaFarooqui: @Kixunil Thanks for the suggestion. Created a new issue #689 to track it.

PR #653, opened 2021-04-13
Changed default password from `changeme` to `password` in doc

ALeschinsky:
Playing around with the latest version on docker hub here: https://hub.docker.com/r/shahanafarooqui/rtl - for the current version 0.10.1 'changeme' doesn't work, but 'password' does, I actually just guessed it as it doesn't seem to be documented anywhere...


ShahanaFarooqui:
Thanks for the PR.

PR #742, opened 2021-07-30
first iteration of enhancedLogger code

violet360:


ShahanaFarooqui (changes requested):
1: File logger.js:
1.1: Config Setup Variable should only being written for debugging purpose and the logLevel has already been set as 'DEBUG'.  So, we can remove the redundant logic to only write in the console if they are not Config Setup Variables (if (msgJSON.fileName !== 'Config Setup Variable') { console.log(msgStr); }).

1.2: For case "DEBUG": we can check if the msgJSON.data is defined within the first if statement itself rather than checking the twice (line 61 & 67)

2: File connect.js:
2.1: Always compare with strict equality (===).
	If we compare null with undefined, it returns true. Also, if(x === undefined) will throw a ReferenceError if the variable in not declared.
	So, checking if(x==undefined) or if(x!==undefined) is not a good practice. You should rather check it like below:
		2.1.1: if(x) or if(!x)
		2.1.2: or if you explicitly want to check the type of the variable as undefined then use if(typeof x === "undefined")

2.2: Can we enhance the logByLevel method, so that it will open and update all logLevels in a single run? Currently, If the user has 5 node setup, then the file will be opened and updated 5 times.

ShahanaFarooqui (changes requested):
1: File logger.js:
1.1: For WARN & INFO cases: console.log messages in else conditions are against the requirement (lines 38-40 & 48-50).

2: File connect.js:
2.1: 'enableLogging' is being used at total 12 places in the file. Nine of them should be replace with the new key.
2.2: Why do we need const config from rxjs (line 14)? It can be removed if we are not using it. Also, if it has to be used, it should not be used as 'config' to avoid context variable errors.
2.3: A better place to call logByLevel method is before connect.validateNodeConfig(config) (line 600). This will ensure that the change will reflect everywhere without writing any extra code and it will be easier to deprecate the method after some releases.
2.4: detectConfigChange(allNodes) method can be replaced with a flag variable, initially set to false and update it to true if node.Settings.hasOwnProperty('enableLogging'). For example:
	var flgUpdateConfig = false;
    config.nodes.forEach(node => {
      if (node.Settings.hasOwnProperty('enableLogging')) {
        flgUpdateConfig = true;
		...
		
3: File RTLConf.js:
3.1: It will still search for the enableLogging key value and can introduce bugs in the future.

4: Docs files:
4.1: Application_configurations, Sample-RTL-Config.json and other .md files still show the enableLogging as valid key. It will cause confusion for the user and RTL will also be unable to deprecate the key and related logic.

ShahanaFarooqui (approved):

PR #745, opened 2021-08-01
Enable password autocomplete

brentkearney:
Modern browsers/OS's can suggest unique, secure passwords and encrypt them for storage. This saves time setting up new passwords and logging in, prevents typing mistakes, and discourages the poor security practice of password reuse. All that is required is the addition of an HTML tag to password fields.


.gitignore
```diff
@@ -53,3 +53,4 @@ RTL-1.conf
 RTL-Multi-Node-Conf-1.json
 RTL-Config-for-BTC-Testing.json
 ECLDummyData.log
+.DS_Store
```
brentkearney, line 56: MacOS file manager leaves these files automatically, irrelevant to RTL.

ShahanaFarooqui:
@brentkearney, Thank you for opening the PR but we decided to not to merge these changes into the application due to security reasons.

We do not encourage our users to use autocomplete and stored passwords. Although it is a bit inconvenient to setup a unique and secure passwords by themselves, it is highly advised.

We also strongly suggest our users to type their passwords every time they want to use the application rather than saving it in the browser's storage.

brentkearney:
OK.

However, most users will end up re-using the same password they use elsewhere, which is a big security risk, and will likely use weaker passwords than the long, random, and unique ones that the browser would generate. Users may also not store their password securely (sticky notes, plaintext files), whereas browsers/OSes/password manager apps encrypt password data. Keystroke loggers will capture typed-in passwords, but perhaps not browser auto-fills.

In any case, after submitting this PR, I discovered that I can evoke Safari's password autofill feature by clicking just above the field by the label, so my modifications are not even needed. They might make it more intuitive though.

🍻

Issue #937, opened 2022-01-27
Hyperlink Channels from the dashboard to the channels grid


saubyk:
The `Channel Capacity` panel on the operator dashboard and the `In-Bound Liquidity` and `Out-Bound Liquidity` on the merchant dashboard have channel listings. Each channel has a unique channel ID, which can searched on the `Peers/Channels` page.

The channels in the panels can be hyperlinked such that user can click on the channel on the dashboard and land on the `Peers/Channel` page, with the channel ID in search. This will help user to perform channel specific actions quickly by navigating from the dashboard directly and save time by not needing to search again on the channels grid.

PR #938, opened 2022-01-30
Channel linking from Dashboard to Open channels page

saiy2k:
PR for https://github.com/Ride-The-Lightning/RTL/issues/937

Tested in LND, C-Lightning, Eclair for both Merchant and Operator layouts.

Fetching the channelId from queryParam and applying filter is done
 * in 'ngOnInit' for c-lightning and eclair.
 * in 'ngAfterViewInit' for LND, as it's not happening right when done in 'ngOnInit'

### DIFF
--- a/src/app/clightning/home/channel-capacity-info/channel-capacity-info.component.html
+++ b/src/app/clightning/home/channel-capacity-info/channel-capacity-info.component.html
@@ -15,7 +15,9 @@
   <div class="channels-capacity-scroll" [perfectScrollbar]>
     <div fxLayout="column" fxFlex="100" *ngIf="activeChannels && activeChannels.length > 0; else noChannelBlock">
       <div *ngFor="let channel of activeChannels" class="mt-2">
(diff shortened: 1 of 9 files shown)


ShahanaFarooqui (changes requested):
Hi Sai,

Can you please change the way we pass the data to filter? Query params make the route too ugly in the browser :) . More importantly, it is not aligned with the current design; we are only passing static route as route params till now. Something like below:

Sending Component:
TypeScript => this.router.navigate(['/xyz'], { state: { name: 'abc' }});
HTML =>	 <a [routerLink]="['xyz']" [state]="{ name: 'abc'}">Link</a>

Receiving Component:
this.routeState = this.router.getCurrentNavigation().extras.state;

ShahanaFarooqui:
@saiy2k Thank you, merged these changes with PR #946 in `Release-0.12.2` branch. We usually keep `master` branch stable because some of our users directly pull the code from master. We create feature branch to work on a feature, then merge that branch into the release branch. Finally, once all the issues from that release branch are closed, we merge the release branch into master and release the latest RTL version.

ShahanaFarooqui:
@saiy2k One more update regarding this PR. I updated the state's key from `channelId` to `filter` to make it reusable for other filtering values as well. For example: Issue https://github.com/Ride-The-Lightning/RTL/issues/942 also needed to redirect to channels page but the data will be filtered by `inactive` status. 

Please refer PR #949 for more details.

PR #941, opened 2022-02-06
Fixed wrong requirement

bota87:


saubyk (commented):
@bota87 
RTL_SSO is an optional configuration. Can you please explain, why you are marking it as 'Required'?

saubyk:
> @bota87
> RTL_SSO is an optional configuration. Can you please explain, why you are marking it as 'Required'?

So @ShahanaFarooqui corrected me. It is actually a 'Required' flag. 😅

saubyk (approved):

PR #979, opened 2022-03-20
Build Docker containers on GitHub actions

AaronDewes:
This builds Docker containers from one Dockerfile, on GitHub actions, to the GitHub container registry.

It uses buildx and the built image will automatically run on arm64, amd64 and armv7 (3 images get pushed and then commpiled into one manifest for all architectures, so every arch can download the right one).

This shouldn't have any impact on anyone using the images, except instead of putting `image: shahanafarooqui/rtl` into docker-compose.yml, it's now `image: ghcr.io/ride-the-lightning/rtl`.


ShahanaFarooqui:
@AaronDewes Thank you for the PR. It will be really useful for the project, but it can affect RTL's integration for multiple providers like BTCPayServer, Umbrel, etc. We are keeping the PR open until we are not done with in-depth impact analysis.

AaronDewes:
> it can affect RTL's integration for multiple providers like BTCPayServer, Umbrel, etc. 

There should be no big issues here, the only required change is a one-time change of one line, and existing users won't have any issues, because it'll only affect future versions.

But please let me know if you see any other impact this could have. I'm developing an umbrel alternative @runcitadel, and our app system is docker-compose based, so I have some experience with this and there should be no big effort to migrate for external developers by doing this.

(Literally just replacing `image: shahanafarooqui/rtl` with `image: ghcr.io/ride-the-lightning/rtl` once the new version is out)

AaronDewes:
For BTCPay, it has a slightly more complicated build system, but it's not that hard to implement either: https://github.com/AaronDewes/btcpayserver-docker/commit/a787f31a2dd9b5e5ddb04b985b6cb71b92dd72d7

saubyk:
Hi @AaronDewes thanks for the PR.

Ideally we should generate the images both via Github actions and Circle CI and give the node solution providers time to transition to Github actions eventually. So basically we will be deprecating Circle CI/Docker Hub images and eventually phase it out.

AaronDewes:
> Ideally we should generate the images both via Github actions and Circle CI and give the node solution providers time to transition to Github actions eventually. 

I don't really understand why the node providers should need transition time, it's a one-time one-line change needed and the old images will still exists, so it's only one step when upgrading.

However, it's also possible to push to Docker hub from GitHub actions, so the new, simpler Dockerfile would be used and only one ci provider, and the old images would still get updated.

ShahanaFarooqui:
@AaronDewes I tried to deploy the build on `ghcr.io/ride-the-lightning/rtl` but I am unable to do it. 

I have created a branch named `ci-test` and updated the push condition on this branch as well. But I am still unable to successfully do that.

AaronDewes:
Weird, because the builds seem to work sucesfully. I'll try in my fork again

ShahanaFarooqui:
Closing it in favour of PR #1421.

PR #1039, opened 2022-06-07
[LND] Add advanced channel policy min/max_htlc_msat

fvdpol:
This PR extends the 'Update Fee Policy' for a specific channel with an advanced options section that exposes the minimum_htlc_msat and maximum_htlc_msat channel policy parameters. 

From a user's perspective the change is minimal, as the additional parameters are not exposed and only visual difference is the presence of a 'Show Advanced' button at the bottom of the form: 
Screenshot 2022-06-07 at 13 53 17

After clicking the 'Show Advanced' button the layout is changed and the additional fields exposed:
Screenshot 2022-06-07 at 13 53 29

Tested with LND 0.14.2


ShahanaFarooqui:
@fvdpol Thank you for the PR.

Please note that I updated the logic to keep min_htlc and max_htlc optional with commit `89e7ebadaa652845779934dd6e0848c8d1842a7e` as these are not required to be sent with every update request.

Issue #1050, opened 2022-07-19
Allow alternate location for log file


jpentland:
**Is your feature request related to a problem? Please describe.**
It does not always make sense for log files to be in the same location as the config files. E.g. if the config file were to be placed at /etc/RTL/RTL-Config.json.

**Describe the solution you'd like**
Allow overriding the log directory location, either via environment variable, command line flag, or somewhere in the config file.

**Additional context**
Config file location is currently set via the environment variable RTL_CONFIG_PATH and is used for both the config file path and log file path.

ShahanaFarooqui:
Steps to make log file path configurable:
- Add `logFilePath` in RTL-Config.json node's setting.
- This `logFilePath` setting should be configurable/editable through the UI.
- Add `LOG_FILE_PATH` in environment variable.
- Environment variable should take precedence over config file.
- Maintain current path as default log path.
- If RTL fails to gain write access on the folder, show warning message to the user and fallback to default logging path.

saubyk:
> * This `logFilePath` setting should be configurable/editable through the UI.

we should be careful in making the `logFilePath` configurable via the UI, lest it opens up the app to LFI vulnerability

ShahanaFarooqui:
UI should not use file explorer for configuration/update. It should be a text input field.

PR #1051, opened 2022-07-22
config.ts: Allow setting log file location with RTL_LOG_PATH

jpentland:
This preserves the previous behavior of using the conf file path if
RTL_LOG_PATH is not set.

Solves: #1050


ShahanaFarooqui:
@jpentland Thank you for the PR. But, we will not be able to merge it in its current state. 

I have updated all necessary steps on the issue. Please make sure that all of them are covered and tested locally. Also note that code linting is mandatory before merging.

I am closing this PR now. Please feel free to open another with all steps covered.

PR #1070, opened 2022-08-21
Move dev-only deps to `devDependencies`

erikarvstedt:
This shrinks the `node_modules` size by 507 MB (660 MB -> 153 MB).

I only checked the largest deps if they are applicable to be moved to `devDependencies`. There might be other, smaller deps that can also be moved.


ShahanaFarooqui:
@erikarvstedt Thank you for the PR. Angular dependencies used to be mandatory before. Did it change after `Ivy`?

PR #1099, opened 2022-09-20
add autocomplete for totp

617a7a:


ShahanaFarooqui:
@617a7a, Thank you for opening the PR but we decided to not to merge these changes into the application due to security reasons.

We do not encourage our users to use autocomplete and store password/token. It is a bit inconvenient to setup a unique and secure passwords by themselves, it is highly advised. We also strongly suggest our users to type their passwords every time they want to use the application rather than saving it in the browser's storage.

2FA Token expires every 30 seconds. 2FA can be disabled through settings, if not required, rather than keeping it enabled and auto filling it afterwards..

PR #1123, opened 2022-10-27
Update with instructions on how to use Apache2

NateNate60:
Some users prefer to use Apache2 instead of Nginx as their reverse proxy (because some people already have Apache2 up for some other reason). This adds a guide on how to do that.


ShahanaFarooqui:
@NateNate60 Thanks for the PR. I rearranged instructions a little bit. Merging the PR in Release-0.13.3 branch.

PR #1136, opened 2022-11-20
chore: npm audit fix

abhiShandy:
Fixes 1 high, 1 critical vulnerability


ShahanaFarooqui:
Hi @abhiShandy thanks for the PR, but we do not accept dependency update PRs due to security reasons. Also, we update dependencies with every release which fixes most of the vulnerability issues. So when you update to the latest version, you'll get the vulnerability fixes along with it.

Closing the PR now.

PR #1186, opened 2023-01-29
Add Connect Action to Addresses

royalpinto007:
Implemented the Connect action on the Addresses grid for better functionality.


ShahanaFarooqui (commented):
This PR is just adding another button to copy the URI. Closing the pr now as 1- adding duplicate action button is not needed, 2- It is confusing for the user to click on a button 'Connect' and getting the text copied instead.

Issue #1177, opened 2023-01-20
Database location should be configurable


ShahanaFarooqui:
Fort nix doesn't allow creating files in the install directory because /nix/store isn't accessible for the rtl process. For security reasons they only allow rtl to access it's data directory (/var/lib/rtl). So database directory also needs to be configurable.

Reference Issue: https://github.com/fort-nix/nix-bitcoin/issues/584

erikarvstedt:
Wouldn't the data dir be the obvious location for this database?

The following would be helpful for automatic deployment schemes, like nix-bitcoin: 
- All RTL settings should be specifiable in a config file (like `RTL-Config.json`).
  This base config file should be treated as read-only by RTL.
- When the user interactively changes settings in the GUI, all differences to the base config should be stored in another file (or database) in the data dir.

ShahanaFarooqui:
@erikarvstedt Thank you for the suggestion. I am closing this issue by making db location configurable for now. We will create another issue with your suggestion after in depth analysis and detailed discussion with @saubyk.

Closing with PR #1199.

PR #1199, opened 2023-02-18
Make db location configurable #1177

ShahanaFarooqui:
Make db location configurable #1177


jonasnick (commented):
Works for me. Thanks!

PR #1335, opened 2023-12-06
Fix Side Navigation issues (toggle and on refresh)

lacksfish:
Toggling the sidenav had some issues (on mobile and desktop view), which I believe these changes will fix.

For example:
* The mobile view considers the sidenav to not be visible initially, although the default flag is set to true. Therefore the menu button needs to be clicked twice for the menu to appear.
* In the desktop view, upon refreshing the page, the sidebar becomes visible; however, the associated content does not reposition correctly and remains overlaid by the sidebar instead of being properly moved to the side.
* Added backdroup-clicked event handler based on Angular documentation

I also took the liberty to add a watch command (`npm run watchfrontenddev`) for the frontend to the `package.json`

This PR does not include the associated production build for clarity of code changes.

## Summary by CodeRabbit

- **New Features**
	- Implemented a backdrop click feature for the side navigation panel.
- **Bug Fixes**
	- Fixed an issue where the side navigation state was not toggled correctly on certain events.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "start": "ng serve --open",
     "prebuildfrontendtest": "node src/prebuild.cjs",
     "prebuildfrontend": "node src/prebuild.cjs",
+    "watchfrontenddev": "ng build --configuration development --optimization false --watch",
     "buildfrontendtest": "ng test --watch=false && ng build",
     "buildfrontend": "ng build --configuration production",
     "buildbackend": "tsc --project ./server/tsconfig.server.json",
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -159,7 +159,10 @@ export class AppComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngAfterViewInit() {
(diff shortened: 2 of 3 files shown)


saubyk:
@coderabbitai review

package.json
ShahanaFarooqui, line 1: Why do we need this script? We already have `start` script for development environment? But it can be added, if it helps other users.

src/app/app.component.ts
```diff
@@ -159,7 +159,10 @@ export class AppComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngAfterViewInit() {
-    if (this.smallScreen || !this.flgLoggedIn) { this.sideNavigation.close(); }
+    if (this.smallScreen || !this.flgLoggedIn) {
+      this.flgSideNavOpened = !this.flgSideNavOpened;
```
ShahanaFarooqui, line 163: This flag toggling hides the side navigation on non-small screens on initial page load. So, had to delete it from here.

PR #1349, opened 2024-02-01
fix: add missing boltz state `transaction.lockupFailed`

jackstar12:
The `transaction.lockupFailed` state is missing in the boltz service implementations. It can happen due to the user sending too little coins to the lockup address.

Before:
image

After:
image


ShahanaFarooqui:
@jackstar12 Thanks for the PR. I re-based your branch on Release-0.15.1 and then built the front-end before merging.

PR #1352, opened 2024-02-05
feat: add boltz service to cln

jackstar12:
The new release of `boltz-lnd`, `boltz-client` supports CLN, so the Boltz service can be shown for CLN aswell.
https://github.com/BoltzExchange/boltz-client/releases

- Added Boltz service in side navigation for CLN
- Added Boltz Service settings in node config => settings tab.


ShahanaFarooqui:
@jackstar12 Did you test it locally? Seems like it should need more changes than these.

jackstar12:
@ShahanaFarooqui yes I did. It doesnt need more changes since the api is the same for both lnd and cln and backwards compatible.

PR #1490, opened 2025-03-30
feat: boltz swap in refund address

jackstar12:
Require a refund address when creating a swap in and paying it externally to make sure the swap can be refunded automatically if it fails.

## Summary by CodeRabbit

- **New Features**
	- Introduced a conditional "Refund Address" input within the swap process. When not using the internal wallet option, users are now prompted to provide a refund address with clear instructions and validation, ensuring a smoother transaction experience.


ShahanaFarooqui (approved):
Fixed lint errors and added compiled builds.

PR #1506, opened 2025-09-09
Release 0.15.6

ShahanaFarooqui:


s373nZ:
@ShahanaFarooqui FYI - after reviewing the security changes, if the install command changed to use `npm ci ...`, then it looks like your release notes template for the Github releases description also needs an update. It still uses `npm install ...`

https://github.com/Ride-The-Lightning/RTL/releases/tag/v0.15.6

ShahanaFarooqui:
> @ShahanaFarooqui FYI - after reviewing the security changes, if the install command changed to use `npm ci ...`, then it looks like your release notes template for the Github releases description also needs an update. It still uses `npm install ...`

Thanks for bringing it to my attention. Updated the text in release notes.

Issue #1351, opened 2024-02-03
NPM install fails with severity vulnerabilities - No fix available


Boondocker4:
**Describe the bug**
When I try to install RTL using npm i get several severity vulnerabilities that are unable to be fixed.  Result is RTL installation fails.

**To Reproduce**
Install nodejs using 
curl -fsSL https://deb.nodesource.com/setup_21.x | sudo -E bash - &&\
sudo apt-get install -y nodejs

wget https://github.com/Ride-The-Lightning/RTL/archive/refs/tags/v0.14.1.tar.gz
tar -xvf v0.14.1.tar.gz
cd RTL-0.14.1
npm install --omit=dev --legacy-peer-deps

**Screenshots**
satoshi@nodebox:~/RTL$ npm install --omit=dev --legacy-peer-deps
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.

changed 1 package, and audited 390 packages in 2s

27 packages are looking for funding
  run `npm fund` for details

2 moderate severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.

Run `npm audit` for details.

satoshi@nodebox:~/RTL$ npm audit --omit=dev --legacy-peer-deps
# npm audit report

request  *
Severity: moderate
Server-Side Request Forgery in Request - https://github.com/advisories/GHSA-p8p7-x288-28g6
Depends on vulnerable versions of tough-cookie
No fix available
node_modules/request

tough-cookie  <4.1.3
Severity: moderate
tough-cookie Prototype Pollution vulnerability - https://github.com/advisories/GHSA-72xf-g2v4-qvf3
No fix available
node_modules/request/node_modules/tough-cookie

2 moderate severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.

**Your environment**) 
* Version of `RTL` 0.14.1-beta
* Version of `lnd` lncli version 0.17.3-beta commit=lightning-terminal-v0.12.2-alpha
* Version of `bitcoind` 26.0.0
* Browser & browser version Brave v 1.61.116
* Operating system Linux nodebox 5.15.0-92-generic #102-Ubuntu SMP Wed Jan 10 09:33:48 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux (Ubuntu 22.04.3 LTS)

* any other relevant environment details

[...]

PR #1512, opened 2025-09-28
fix: override request module with @cypress/request

3nprob:
The `request` module is since long deprecated, abandoned by the maintainer, and has known security issues (yes, including ones legitimately relevant for RTL usage).

While migrating it and equally deprecated `request-promise` to `fetch` might be even better, an interim solution would be to switch it to the maintained and by now established `@cypress/request` fork which addresses known bugs.
- https://www.npmjs.com/package/@cypress/request
- https://github.com/cypress-io/request

Related:
- #1351
- https://github.com/Ride-The-Lightning/RTL/discussions/1147
- https://github.com/Ride-The-Lightning/RTL/discussions/1122

---

@coderabbitai ignore

(iffy privacy policy)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "otplib": "12.0.1",
         "pdfmake": "0.2.10",
         "process": "0.11.10",
-        "request": "2.88.2",
+        "request": "npm:@cypress/request@3.0.9",
         "request-promise": "4.2.6",
         "rxjs": "7.8.1",
         "sha256": "0.2.0",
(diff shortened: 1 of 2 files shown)


3nprob:
@saubyk @ShahanaFarooqui PTAL

ShahanaFarooqui:
@3nprob Thanks for opening the PR! I’d prefer to migrate to `axios` instead of using `@cypress/request`. However, the migration is turning out to be more time-consuming than expected. I’ve created a new branch, `update-request-to-axios`, with partial progress — the URL handling, request parameters, and query parameters still need to be adjusted. More importantly, the changes will require thorough testing. So the final PR might take a bit longer to complete but I am close this one for now.

PR #1536, opened 2026-01-14
Fix for displaying keysend memo

### LINKED ISSUE #1487: Print keysend message in invoices memo column
Followup of #467, but for keysends

For LND, Invoices memo field is displayed in the invoice tables, but for keysends, the memo column is empty even when keysends have messages

Indeed, when looking up the payment hash of a keysend, the memo field is empty.
The message of the keysend is in htlcs/custom_records/34349334 field of the lookupinvoice command, as a basic hex-string encoding of the keysend text

It would be nice to get those in the memo column too

saubyk:
Fixes #1487


ShahanaFarooqui:
Closing this in favour of PR #1539

PR #1582, opened 2026-03-26
Add Disable Authentication option

### LINKED ISSUE #1285: Single Sign On for Umbrel
Add SSO feature for Umbrel.

ShahanaFarooqui:
Disable Authentication option added for advanced users:
- This option would make sense for node platform vendors who want to provide their own authentication mechanism over the application's.

Non vendor users shouldn't be using this option at all.


.github/docs/Application_configurations.md
```diff
@@ -56,6 +57,7 @@ PORT (port number for the rtl node server, default 3000, Optional)<br />
 HOST (host for the rtl node server, default localhost, Optional)<br />
 DB_DIRECTORY_PATH (Path for the folder where rtl database file should be saved, default RTL root directory, Optional)
 APP_PASSWORD (Plaintext password to be provided by the parent container, NOT suggested for standalone RTL applications, to be used by Umbrel) (Optional)<br />
+DISABLE_AUTH (Flag to disable authentication, NOT suggested for standalone RTL applications, to be used by Umbrel) (Optional)<br /
```
saubyk: Don't think we should specifically call out Umbrel here, it can be used by any vendor.

Recommendation:
Flag to disable authentication, **NOT** recommended for standalone RTL applications, only to be used by Vendors providing their own authentication service

.github/README.md
```diff
@@ -124,6 +124,8 @@ For details on all the configuration options refer to [this page](./docs/Applica
 RTL requires the user to be authenticated by the application first, before allowing access to LND functions.
 Specific password must be provided in RTL-Config.json (in plain text) for authentication. Password should be set with `multiPass:<user defined>` in the `Authentication` section of RTL-Config.json. Default initial password is `password`.
 
+For hosted solutions such as BTCPayServer, we implemented an "SSO" setup using a one-time-use cookie. For Umbrel, we introduced a "disableAuth" option, which disables authentication at the RTL level. This remains secure on Umbrel because only authenticated Umbrel users are granted access to RTL; unauthenticated users are redirected to Umbrel’s login page.
```
saubyk: same thing here, don't think it makes sense to call out Umbrel.

Suggestion:
For hosted solutions such as BTCPayServer, we implemented an "SSO" setup using a one-time-use cookie. For other vendors which have their own authentication service, we introduced a "disableAuth" option, which disables authentication at the RTL level. When using this option, the authentication security is the responsibility of the Vendor. This option is **NOT** recommended for standalone users of RTL.

saubyk (approved):
Just some documentation recommendations.

saubyk (approved):

PR #1622, opened 2026-07-16
Add baseline screenshot capture, fix fixture's blacklisted password

saubyk:
Adds the baseline screenshot harness, and fixes a bug in the fixture merged in #1621.

## Bug: the fixture's password is blacklisted by RTL

`PASSWORD_BLACKLIST` in `src/app/shared/services/consts-enums-functions.ts:302` is `['password', 'changeme', 'moneyprintergobrrr']`. When a blacklisted password is used, `login.component.ts:87` sets `defaultPassword: true` and `app.component.ts:97` redirects to `/rtl/settings/auth` to force a password change.

The fixture set `multiPass` to exactly `"password"` — the first entry on that list. So every fresh browser session logged in, got bounced to the change-password screen, and never reached the dashboard. It's session-based, not first-run only, so it happened every time.

Password is now `rtldev`. `.env` documents why it must stay off the blacklist.

## capture/

Drives RTL against the fixture and captures a fixed set of screens for the 1.0 design baseline.

The value is that capture conditions live in code, not in someone's head. Viewport (1440×900), device scale factor (2, so files are 2880×1800), browser and the screen list are pinned in `capture.mjs`. When 2.0 lands, the "after" can be shot under identical conditions — otherwise a re-shoot on a different display silently invalidates every before/after comparison.

```bash
cd docker/capture && npm install && npx playwright install chromium
node capture.mjs empty   # before ./scripts/seed.sh -- empty states
node capture.mjs full    # after  -- populated
```

Screenshots are gitignored; they belong in the RTL-Design repo.

## Routes were verified, not assumed

Angular renders its 404 component **without changing the URL**, so a route can look fine while being a "Page Not Found". My first pass compared `location.pathname` and reported 24/24 OK. Re-probing by page content found three of them are actually 404s:

- `/rtl/settings/config`
- `/rtl/settings/nodesettings`
- `/rtl/settings/pglayout`

Node Config is `/rtl/settings/bconfig`. The script only contains content-verified routes.

## Verified

29 screens captured, 0 skipped, against the seeded fixture. Node switching across alice/bob/carol works. bob's forwarding history shows all 5 seeded routed payments with the exact amounts — the screen a two-node fixture leaves empty.

## Known gaps

- Loop and Boltz aren't configured, so those two shots record the unconfigured state
- Onboarding/first-run can't be captured, since the config pre-configures three nodes
- Night mode and persona switching aren't scripted yet; both are per-node settings

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -24,6 +24,9 @@ ALICE_REST_PORT=8081
 BOB_REST_PORT=8082
 CAROL_REST_PORT=8083
 
-# RTL
+# RTL. Must not be one of RTL's blacklisted weak passwords
+# ('password', 'changeme', 'moneyprintergobrrr') or RTL forces a password change
+# on every login and parks you on the auth settings screen. Set in
+# rtl/RTL-Config.regtest.json as multiPass; this is only used for messages.
 RTL_PORT=3000
-RTL_PASSWORD=password
+RTL_PASSWORD=rtldev
--- a/docker/README.md
+++ b/docker/README.md
@@ -28,7 +28,7 @@ docker compose up -d          # bitcoind, alice, bob, carol, rtl
 ./scripts/seed.sh             # fund, connect, open channels, make payments
 ```
 
-Then open <http://localhost:3000> — password `password`. All three nodes appear in
+Then open <http://localhost:3000> — password `rtldev`. All three nodes appear in
 the node switcher.
 
 Tear down, discarding all state:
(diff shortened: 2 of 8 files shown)


saubyk:
Splitting this up — it bundled two unrelated changes. The password fix stands alone in #1623. The capture harness belongs in the RTL-Design repo, not here: it doesn't import any RTL code (it just drives http://localhost:3000), nothing in RTL's CI exercises it, and the baseline it produces is an RTL-Design deliverable.

Issue #1634, opened 2026-07-19
Remaining npm audit findings need code changes: replace request/request-promise, csurf, and the crypto-browserify polyfills


saubyk:
After the batch dependency update in #1633, `npm audit --omit=dev` still reports **13 production vulnerabilities (2 critical)**. None of them can be fixed by a version bump — each needs a code change. Filing this to track that work.

## 1. Replace the deprecated `request` / `request-promise` stack — critical

`request` has been deprecated and unmaintained since 2020, with an unfixed SSRF advisory (GHSA-p8p7-x288-28g6). It also pins vulnerable copies of `form-data` (critical, GHSA-fjxv-7rqg-78g4 / GHSA-hmw2-7cc7-3qxx), `qs`, `tough-cookie` and `uuid` — **8 of the 13 remaining findings clear automatically when it goes**.

- Scope: **36 files** in `server/` import `request-promise` — `server/utils/common.ts` plus essentially every CLN and Eclair controller (LND controllers use it too via shared helpers).
- Suggested replacement: `axios`, which is already a production dependency and already used elsewhere in the backend — no new dependency needed. Native `fetch` (Node ≥ 18) is the zero-dependency alternative.
- Main porting concerns: `request-promise` resolves with the parsed body directly and rejects on non-2xx via `statusCodeError`; option names differ (`qs` → `params`, `rejectUnauthorized` → an `https.Agent`); CLN's rune header auth and Eclair's basic auth both need to carry over. The docker regtest fixture (all three implementations) is the natural end-to-end verification.

## 2. Replace deprecated `csurf` — low

`csurf` is deprecated (its advisory chain is via an old `cookie`; npm's only "fix" is a downgrade). Used in one place: `server/utils/authCheck.ts` (`csurfProtection`, cookie mode). Replace with a maintained double-submit-cookie middleware (e.g. `csrf-csrf`) or an in-house implementation of the same pattern. Note the frontend contract: the token is read from the `XSRF-TOKEN` cookie and echoed as the `x-xsrf-token` header, so a drop-in replacement must keep those names.

## 3. `crypto-browserify` polyfill chain — low

`crypto-browserify`, `browserify-sign`, `create-ecdh` and `elliptic` (GHSA-848j-6mx2-7j84, risky-implementation advisory with no fixed release) are browser polyfills for the frontend build. Options, in rough order of preference:

[...]

PR #1643, opened 2026-07-20
Replace deprecated csurf with csrf-csrf

saubyk:
Item 2 of #1634: replaces the deprecated `csurf` middleware (unmaintained since 2022, pins an old `cookie` release with a known advisory) with the maintained **`csrf-csrf` v4** — the double-submit-cookie middleware suggested in the issue.

## Approach

Same signed double-submit pattern csurf implemented, now HMAC-signed and **session-bound**, keyed on the existing boot secret (`common.secret_key`). The three csurf touchpoints map cleanly:

- **`server/utils/csrf.ts`** — configures `doubleCsrf` and mounts `doubleCsrfProtection` globally (still skipped in `NODE_ENV=development`). Config keeps the wire contract identical: signed cookie stays named `_csrf` (now httpOnly; `secure: false` to match the session cookie on plain-HTTP deployments), token sources match csurf's full list (`x-xsrf-token`, `x-csrf-token`, `csrf-token`, `xsrf-token` headers, `body._csrf`, `query._csrf`), and the error code is `EBADCSRFTOKEN` — already handled by `app.ts`'s error handler, so 403 behavior is unchanged.
- **`server/utils/authCheck.ts`** — `verifyWSUser` now routes through the shared middleware instead of its own csurf instance. Upgrade requests are GETs (an ignored method in both libraries), so the handshake semantics are unchanged — and the middleware never touches `res` on that path, so the `null` res stays safe.
- **`server/utils/app.ts`** — no contract change; `doubleCsrfProtection` attaches `req.csrfToken()` so the `XSRF-TOKEN` cookie (Angular) and response header (Quickpay) logic works as before. One real fix: it called `req.csrfToken()` **twice** (cookie + header). Under csurf every generated token validated against a stable per-request secret; under csrf-csrf each first-visit call mints a fresh token, desyncing the `XSRF-TOKEN` cookie from the `_csrf` cookie it must equal (found live on the fixture — first login 403'd). The token is now generated once per request.

**Security upgrade over csurf**: tokens are HMAC-bound to the session identifier, so a token+cookie pair stolen from one session no longer validates under another — csurf's cookie mode performed no such binding.

No frontend changes: Angular still reads the `XSRF-TOKEN` cookie and echoes `x-xsrf-token`; Quickpay still reads the `XSRF-TOKEN` response header.

## Audit impact

`npm audit --omit=dev`: **6 low → 4 low**. `csurf` and its stale `cookie` pin are gone; the remainder is the `crypto-browserify`/`elliptic` polyfill chain (item 3 of #1634). `csrf-csrf` is pinned exactly (4.0.3) per repo convention.

## Testing

Against the docker regtest fixture (branch image):

- **Both API suites re-pass** — 43 checks across LND, Core Lightning and Eclair (every one performs the CSRF handshake and authenticated POSTs, so the middleware is exercised on every call).
- **Dedicated CSRF battery (8 checks)**: valid-token authenticate; missing token → 403; garbage token → 403; **cross-session token replay → 403** (the new session binding); token stable across requests within a session; `XSRF-TOKEN` cookie === `_csrf` cookie; `XSRF-TOKEN` response header present (Quickpay contract); websocket handshake accepted through the rewired `verifyWSUser`.
- Lint and backend build clean.


saubyk:
## Review

Verified the migration independently, including reading the `csrf-csrf@4.0.3` source and running a full CSRF lifecycle test against a local instance (`NODE_ENV=production`, dummy node). The wiring is correct — but there is one should-fix on the logout → re-login flow, posted as a line comment on the session binding, plus two minor notes below.

Checked:

- **Library semantics** (from source): error code is `EBADCSRFTOKEN` → `app.ts`'s 403 handler applies; ignored methods are `GET/HEAD/OPTIONS`, so the websocket upgrade through `verifyWSUser` never touches the `null` res; validation is strict cookie↔token equality **plus** HMAC over the session id — the session binding is real, and an invalid cookie token is silently re-minted on the next `req.csrfToken()` call (`validateOnReuse: false` default), so reloads self-heal.
- **The once-per-request fix in `app.ts` is necessary and correct**: each `req.csrfToken()` call on a first visit mints a fresh token *and* re-sets the `_csrf` cookie (last `Set-Cookie` wins), so the old double call genuinely desynced cookie from header.
- **Middleware order**: `saveUninitialized: true` gives a stable session id from the first response; session → cookieParser → CSRF → routes, so `req.session.id` and `req.cookies` are populated where needed. Login never regenerates the session (`authenticate.ts`), so no post-login desync.
- **Token sources** match csurf's full list; lockfile fully sheds the csurf chain (`csrf`, `rndm`, `tsscmp`, old `cookie` pins); exact pin per convention.
- **Empirically**: handshake sets `_csrf` == `XSRF-TOKEN` cookie == header; garbage token → 403; valid login → 200; logout → 200; fresh page load rotates the token → login 200.
- Lint clean; `npm run buildbackend` reproduces the committed `backend/` artifacts exactly; release note follows the established format.

### Minor: session cookie is now mandatory for API consumers

Under csurf a script could handshake once (`GET /`) and then POST with only the `_csrf` cookie + token header. With session binding, dropping `connect.sid` → 403. That's the advertised tightening, but worth one line in the release notes for third-party script/API users.

### Nit: static-served paths no longer set `_csrf`

csurf set the `_csrf` cookie on every request with an ignored method, including static-served `GET /rtl/`; csrf-csrf's middleware is passive on ignored methods, so only the catch-all (`GET /` or any non-static deep link) sets the cookies. No SPA impact (XSRF-TOKEN was only ever set by the catch-all), but scripts handshaking against `/rtl/` directly now receive no CSRF cookie. Related edge: `_csrf` has no `maxAge` while `connect.sid` expires after a day — after session-cookie expiry, the same 403-on-login symptom as the line-comment finding until the page is refreshed.

server/utils/csrf.ts
```diff
@@ -1,14 +1,31 @@
...
 
+  // Signed double-submit-cookie protection (replaces the deprecated csurf).
+  // The signed token lives in the httpOnly '_csrf' cookie; the client echoes
+  // the same token (read from the XSRF-TOKEN cookie set in app.ts) in a
+  // header. The cookie is not secure-only because RTL commonly serves plain
+  // HTTP (matching the session cookie); token sources match what csurf
+  // accepted. The error code EBADCSRFTOKEN is handled in app.ts.
+  private doubleCsrfUtilities = doubleCsrf({
+    getSecret: () => this.common.secret_key,
+    getSessionIdentifier: (req: any) => (req.session ? req.session.id : ''),
```
saubyk, line 19: **Should fix:** this session binding breaks the logout → re-login flow until a manual page refresh.

`logoutUser` (`server/controllers/shared/authenticate.ts`) destroys the session and clears `connect.sid`, but the `_csrf` cookie (no `maxAge`) survives, still bound to the destroyed session id. The frontend logs out via `router.navigate(['./login'])` (`rtl.effects.ts`) — no document reload — so nothing re-mints the token. The re-login POST then creates a fresh session (new id) → HMAC mismatch → 403 "Invalid CSRF token, form tempered."

Reproduced live on a local instance (`NODE_ENV=production`): handshake → login 200 → logout 200 (`connect.sid` cleared, `_csrf` kept) → re-login with the same token → **403**, and retrying stays 403 (no self-heal); only a full page reload rotates the token and lets login through. This hits manual logout *and* the idle-timeout auto-logout (`angular-user-idle`) — a core flow that worked under csurf because the token wasn't session-bound.

Candidate fixes, in order of preference:

1. **Frontend**: full document navigation on logout — the same effect already does `window.location.href = ...` for the SSO path; the fresh GET re-mints the token. One-liner, but breaks the "no frontend changes" claim.
2. **Backend**: re-mint `_csrf`/`XSRF-TOKEN` on the `EBADCSRFTOKEN` error path so a retry succeeds (the first attempt still fails with the scary message — partial fix).
3. **Backend**: keep `connect.sid` on logout (destroy the session data only) — express-session reuses the incoming signed session id on a store miss, so the existing token keeps validating. Zero UX change, slightly weaker session hygiene.

saubyk: Fixed in cd6ce741 with your option 1 plus option 2 as a backstop:

**Option 1 (primary)**: the logout effect now does a full document navigation to the login page — but *after* the server logout completes rather than before, so the reload can't abort the `/logout` request (the old code raced them; a `catchError` still navigates if the call fails). One wrinkle you couldn't see from the backend: the logout reason travels on the NgRx action stream (`login.component.ts` subscribes to `LOGOUT`), which no full-reload variant can survive — so the reason is handed across the reload via sessionStorage (written after `clearAll()`, picked up and cleared by the login component). The SSO branch already left the document and is unchanged.

**Option 2 (backstop)**: the `EBADCSRFTOKEN` error path now re-mints the token for the current session before responding, so any client holding a stale token self-heals on retry — this also covers the server-restart case (boot-secret rotation invalidates tokens the same way), which option 1 alone wouldn't.

Verified on the fixture: your exact repro now runs login 200 → logout 200 → stale-token login 403 (binding intact) **with re-minted `XSRF-TOKEN`/`_csrf` cookies on the 403** → retry 200; and the reload path (fresh `GET /` after logout) logs in on the first attempt. Both API suites (43 checks), the 8-check CSRF battery, and the full frontend suite (199 specs) pass; frontend + backend artifacts rebuilt.

saubyk:
Both notes addressed in 91924b64 — thanks for the independent verification (and for confirming the once-per-request fix and the lockfile shed).

**Session cookie now mandatory**: added to the release-note entry as requested — API consumers must carry `connect.sid` alongside `_csrf` and the token header, since a token without its session no longer validates.

**Static-path handshake + `_csrf`/`connect.sid` lifetime skew**: documented in the same note — scripts should handshake against `GET /` (or any non-static route), and a stale token (destroyed/expired session, or an RTL restart rotating the boot secret) now self-heals: the `EBADCSRFTOKEN` 403 response re-mints fresh `XSRF-TOKEN`/`_csrf` cookies, so a retry succeeds (added in cd6ce741 for the logout finding; it covers the session-expiry skew the same way). For the in-app case the expiry edge is additionally covered by the idle-timer logout, which now does a full document navigation and re-handshakes. Chose to document rather than re-mint on every safe request — keeping token-minting off the static path keeps the change surface of a security PR small, but happy to add a refresher middleware for exact csurf ergonomics if you'd prefer.

saubyk:
Follow-up on integration impact: verified the **BTCPayServer** flow specifically, since btcpayserver-docker runs RTL in SSO mode (`RTL_SSO: 1`) with entry via `/rtl/api/authenticate/cookie?access-key=…` and logout redirecting to `/server/services`.

- That entry URL is not a registered route, so it falls to the token-minting catch-all — the SPA boots with valid CSRF cookies and SSO login succeeds. **A/B-verified against both the csurf build and this branch: identical behavior (login 200).**
- The logout change only touches the non-SSO branch; BTCPay's SSO logout (full navigation to `logoutRedirectLink`) is unchanged.
- `sameSite: strict` is a non-issue behind BTCPay's same-host Traefik path routing, and `secure: false` matches the session cookie behind the HTTPS proxy.
- One pre-existing quirk surfaced by the A/B (not a regression): entering literally at `/rtl/` leaves the SPA without an `XSRF-TOKEN` cookie and the first login 403s **on the csurf build too**; this branch improves it — the 403 re-mints cookies so a retry succeeds.

BTCPay currently pins `rtl:v0.15.4`, so nothing reaches those deployments until they bump versions.

PR #1644, opened 2026-07-20
Drop crypto-browserify polyfills by moving 2FA TOTP to WebCrypto

saubyk:
Final item of #1634 — removes the last production `npm audit` findings by dropping the `crypto-browserify` polyfill chain. **`npm audit --omit=dev` now reports zero vulnerabilities** (down from 13, including 2 critical, at the start of this cleanup series: #1638 → #1643 → this).

## Root cause

The frontend build mapped `crypto` → `crypto-browserify`, `stream` → `stream-browserify`, `vm` → `vm-browserify` in `tsconfig.json` `paths`. Deleting those mappings and rebuilding pinpointed the sole consumer: **`otplib`** (`@otplib/plugin-crypto` does `require('crypto')`), used only by the two-factor-auth settings dialog. That polyfill chain carried every remaining finding — the `elliptic` advisory (GHSA-848j-6mx2-7j84, *no fixed release*, so unbumpable) plus `browserify-sign`/`create-ecdh`.

## Change

- New `src/app/shared/services/totp.service.ts` — RFC 6238 TOTP (HMAC-SHA1, 6 digits, 30s step, window 0) on **WebCrypto** (`crypto.subtle`), plus base32 encode/decode and a `keyuri` builder.
- `two-factor-auth.component.ts` (the only browser `otplib` consumer) uses it instead of `otplib`. `check()` is now async — WebCrypto's digest is promise-based — so the verify handler was moved into a `.then()`; its return value was never used for control flow (plain click handler), so no behavior change.
- Removed `otplib` from the bundle and `crypto-browserify`/`stream-browserify`/`vm-browserify` + their `tsconfig` `paths` from the tree. `buffer`/`process` stay (still used by `polyfills.ts`, not vulnerable).

**The backend still verifies login tokens with `otplib`** (`server/controllers/shared/authenticate.ts`, unchanged), so the new client-side algorithm has to match it exactly. It does — verified byte-for-byte against `otplib` and the RFC 6238 appendix B vectors. Existing authenticator app enrollments keep working with no migration.

## Testing

- **Unit spec** (`totp.service.spec.ts`, 5 tests): RFC 6238 SHA1 vectors at the standard epochs, `check()` accept/reject incl. non-numeric/empty, invalid-base32 handling, secret shape, and exact `keyuri` parity with `otplib.authenticator.keyuri`.
- **E2E against the docker fixture**: generated a 2FA secret with the new WebCrypto service, confirmed a token it produces is **accepted by the backend's `otplib`** at login, and that wrong/absent tokens behave as before; then disabled 2FA to restore. Also asserted `webcrypto.generate(secret) === otplib.generate(secret)` on a freshly generated secret.
- Both API suites (43 checks) pass; full frontend suite **204 specs** (199 + 5 new); lint and both production builds clean; confirmed the three polyfill packages are absent from the built container while `otplib` remains for the backend.


saubyk (commented):
**Review verdict: approve.** Verified the claims independently rather than taking them at face value — the core claim (exact interoperability with the backend's otplib) holds for every input RTL can actually produce. One non-blocking nit below.

## Verified independently

| Claim | Result |
|---|---|
| `npm audit --omit=dev` zero vulns | ✅ `found 0 vulnerabilities` |
| RFC 6238 vectors match otplib | ✅ Ran otplib at the 4 epochs: `287082 / 081804 / 005924 / 279037` — identical to the spec's expectations |
| Byte-for-byte `keyuri` parity | ✅ Ran `authenticator.keyuri('', 'Ride The Lightning (RTL)', …)` — string-identical, including param order |
| Token parity vs otplib | ✅ Fuzzed 544 (secret, epoch) pairs against otplib: lengths 2–8 and 10–32 bytes all match (480/480), lowercase secrets match |
| `check()` edge cases | ✅ otplib's `isTokenValid` is `/^(\d+)$/` — same regex as the service; 7-digit/non-numeric/empty rejected on both sides |
| Secret padding parity | ✅ Read `@otplib/core` source: `totpPadSecret` repeats-to-20-bytes below 10 bytes — matches `createHmacKey` (one exception, below) |
| Full frontend suite | ✅ **204/204 SUCCESS** |
| New spec | ✅ 5/5 in real headless Chrome |
| Production build | ✅ `ng build` clean, and `git status` is clean afterward — the committed `frontend/` artifacts are byte-identical to a fresh rebuild |
| Bundles actually dropped the chain | ✅ 0 references to `otplib`/`elliptic`/`browserify` in the `main`/`polyfills` bundles and `3rdpartylicenses.txt`; `index.html` ↔ hashed bundles consistent |
| Backend untouched & builds | ✅ `verifyToken` still otplib + window 0; `tsc` clean; `process`/`buffer` correctly retained for `polyfills.ts` |
| Component async refactor | ✅ `onVerifyToken`'s return is unused (plain `(click)`); `flgValidated`/`isTokenValid` semantics preserved; `check()` can never reject (internal `catch → false`), so no unhandled rejections |
| Lint / release note | ✅ ESLint clean; entry correctly filed under Code Health with #1634 cross-ref |

## Finding (non-blocking)

**`createHmacKey`'s "exact parity" comment overpromises — for 1-byte and 9-byte secrets only.** otplib's pad is:

```js
new Array(minLength - currentLength + 1).join(hexSecret) // currentLength = hex chars
```

which under-repeats: for 1-byte and 9-byte secrets it yields an **18-byte** key, while the service always yields 20 bytes. The fuzz confirms: the only mismatches in 544 trials were lengths 1 and 9. Unreachable in RTL — secrets are always 10 bytes from `generateSecret()`, and the secret field is read-only — so zero practical impact, and replicating an otplib bug has negative value. Suggest just softening the comment:

```ts
// Mirrors otplib's totpPadSecret for SHA1 (repeat to 20 bytes below 10 bytes;
// otplib itself produces 18 bytes for 1- and 9-byte secrets — irrelevant here
// since RTL only ever uses the 10-byte secrets generated above).
```

## Nits (take or leave)

- **Double-click window**: verify is now async, so two fast clicks on Verify can dispatch `updateApplicationSettings` twice (was synchronous before). Idempotent payload, harmless.
- `keyuri` hardcodes `period=30&digits=6&algorithm=SHA1` — fine, and the header comment already flags the must-stay-in-sync requirement.

saubyk:
Thanks for the thorough independent verification — the 544-pair fuzz is exactly the check that matters here. Both points addressed in 0253e19a:

**Finding — `createHmacKey` comment**: corrected. You're right that otplib's `totpPadSecret` under-repeats to 18 bytes for 1- and 9-byte secrets where this service pads to 20, so "exact parity" was an overclaim. Since RTL only ever produces 10-byte secrets (and the field is read-only), I took your suggestion and softened the comment to state parity holds for the secrets used here rather than universally — deliberately *not* replicating the otplib quirk, which as you note has negative value.

**Nit — async double-click**: took it. `onVerifyToken`'s check is now async, so I added an in-flight guard (`verifyingToken`) so a second click can't dispatch `updateApplicationSettings` twice — restoring the pre-async single-dispatch behavior even though the payload was idempotent.

Left the `keyuri` hardcoding as-is per your note (the header comment already flags the must-stay-in-sync constraint). TOTP spec (5/5) and lint pass; frontend/backend artifacts rebuilt and committed.

PR #1598, opened 2026-05-12
Prevent multi-node config overwrite during application settings updates

CosimoRicciardi:
## Summary

Fixes an issue where RTL overwrites or drops additional nodes from `RTL-Config.json` when updating application settings in a multi-node setup.

This was reproducible with mixed LND + CLN configurations:

* after changing settings from the UI
* RTL rewrote the config file
* secondary nodes disappeared from the config/UI

## Root Cause

Two behaviors contributed to the issue:

1. `updateNodeSettings()` replaced the entire `node.settings` object:

```js
node.settings = req.body.settings;
```

2. `updateApplicationSettings()` rewrote the full config using only the nodes present in the current request body, causing non-selected nodes to be lost.

### Preserve existing node settings

Changed:

```js
node.settings = req.body.settings;
```

To:

```js
node.settings = { ...node.settings, ...req.body.settings };
```

### Preserve existing nodes during application updates

Before writing the updated config:

* load existing `RTL-Config.json`
* merge back nodes missing from the incoming request payload
* preserve multi-node configurations

## Result

RTL now correctly preserves:

* multiple configured nodes
* mixed LND + CLN environments
* node-specific settings and authentication entries

Fixes #1597

PR #1645, opened 2026-07-20
Prevent multi-node config overwrite during application settings updates

saubyk:
Lands @CosimoRicciardi's multi-node config-overwrite fix (originally #1598) rebased onto the release branch, with a release note and the contributor's regression test. Opened as a fresh PR to avoid force-pushing the contributor's fork branch; their six commits are cherry-picked as-is with authorship preserved.

Closes #1598.

## The bug

On a multi-node setup, saving application settings could corrupt a node's authentication. `CommonService.addSecureData` re-attached each node's secret paths by matching the saved config against the in-memory config **by array position** (`this.appConfig.nodes[i]`). Once nodes were reordered or one was removed, position `i` no longer referred to the same node, so one node's `macaroonPath` / `runePath` could be written onto another.

## The fix (by @CosimoRicciardi)

- Match nodes by `node.index` via a lookup `Map` instead of array position, in both `addSecureData` and the settings-write path (`updateApplicationSettings` / `updateNodeSettings`).
- Merge new settings over existing per-node `settings`/`authentication` rather than wholesale replacing them, guarding missing `authentication` objects.
- Sanitize the persisted `RTL-Config.json` of runtime-only fields before writing — the request `options` object and the resolved `runeValue` — so transient auth material never lands on disk.
- Adds `test/backend/rtlconf.test.mjs` (Node's built-in test runner) asserting indexed auth is preserved and only persisted config is sanitized.

## Provenance & verification

- Six commits cherry-picked from #1598 unchanged (authorship = Cosimo Ricciardi); the only added commit is the release-note entry.
- Rebased onto `Release-0.15.9` with **zero conflicts** despite overlapping the files reworked by the axios (#1638) and csrf (#1643) migrations — the sanitizer's `delete authentication.options` correctly accounts for the axios-era options object.
- `buildbackend` compiles clean, `npm run lint` passes, and the contributor's `rtlconf.test.mjs` passes.

> Note: this is auth-critical config-persistence logic. Recommend a docker-fixture multi-node smoke test (update settings, confirm each node's macaroon/rune survives) before merge, on top of the unit test above.

### DIFF
--- a/backend/controllers/shared/RTLConf.js
+++ b/backend/controllers/shared/RTLConf.js
@@ -200,7 +200,7 @@ export const updateNodeSettings = (req, res, next) => {
     const config = JSON.parse(fs.readFileSync(RTLConfFile, 'utf-8'));
     const node = config.nodes.find((node) => (node.index === req.session.selectedNode.index));
     if (node && node.settings) {
-        node.settings = req.body.settings;
+        node.settings = { ...node.settings, ...req.body.settings };
         if (req.body.authentication.boltzMacaroonPath) {
             node.authentication.boltzMacaroonPath = req.body.authentication.boltzMacaroonPath;
         }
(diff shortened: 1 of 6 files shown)


saubyk:
### Docker-fixture smoke test — differential A/B ✅

Verified this fix end-to-end against the regtest fixture (5 nodes: alice/bob/carol LND, cln CLN, eclair ECL), and A/B'd against a pre-fix `Release-0.15.9` build to prove the test is meaningful rather than a tautology.

**Test:** fetch the config as the frontend does (secrets stripped), **reverse the nodes array**, POST it to `updateApplicationSettings` — the exact input the old position-based `addSecureData` mishandles — then assert each node's persisted auth is preserved *by index*.

| Node (index) | pre-fix `Release-0.15.9` | this PR |
|---|---|---|
| alice (1) | ❌ got eclair's `lnApiPassword` | ✅ keeps alice macaroon |
| bob (2) | ❌ got cln's `runePath` | ✅ keeps bob macaroon |
| carol (3) | ✅ (fixed point of the reversal) | ✅ |
| cln (4) | ❌ got bob's macaroonPath | ✅ keeps cln rune |
| eclair (5) | ❌ got alice's macaroonPath | ✅ keeps eclair password |

Baseline: **6 failed** (all cross-graft in the persisted `RTL-Config.json`). This PR: **18/18 passed**.

**Notable:** on the baseline, live `getinfo` for every node still succeeded *even with the config corrupted* — the in-memory auth was already cached, so the damage is latent in the persisted file and only surfaces on the next RTL restart. That makes the bug more insidious than a live check would reveal, and confirms asserting against the written config (not just runtime auth) is the correct verification.

Also confirmed: no runtime-only fields (`authentication.options`, `authentication.runeValue`) leak into the persisted file, and all five nodes authenticate after the save. Contributor's `test/backend/rtlconf.test.mjs` passes; `buildbackend` + `lint` clean.

PR #1651, opened 2026-07-24
Bound remaining unbounded alias-resolution fan-outs in LND graph.ts and channels.ts   Fixes #1630

### LINKED ISSUE #1630: Bound remaining unbounded alias-resolution fan-outs in LND graph.ts and channels.ts
### Summary

Follow-up to #1501 / #1629. That work bounded the per-peer/per-hop alias resolution fan-out (unbounded `Promise.all(map(getAlias…))`) for the CLN peers + route and the LND peers paths, to stop a large node from firing one alias-lookup request per peer at once (which triggers clnrest `Resource temporarily unavailable (os error 11)` on CLN). The same unbounded pattern still exists in the **LND graph and channels** controllers and should get the same 20-way `runWithConcurrencyLimit` bound for parity.

### Remaining unbounded fan-out sites

**`server/controllers/lnd/graph.ts`**
- `queryRoutes` (~L86): `Promise.all(body.routes[0].hops?.map((hop) => getAliasFromPubkey(...)))` — per-hop alias resolution. Low cardinality (a handful of hops per route), so low impact, but unbounded.
- `getAliasesForPubkeys` (~L144): `Promise.all(pubkeyArr?.map((pubkey) => getAliasFromPubkey(...)))` — cardinality depends on the caller-supplied pubkey list; potentially large.

**`server/controllers/lnd/channels.ts`**
- Channel list (~L34) and pending channels (open/force-closing/closing/waiting-close aggregated into `promises`, ~L72–83) and the second list (~L105) all resolve aliases via unbounded `Promise.all(map(getAliasForChannel…))`. These are the LND analogue of the CLN channel-list fan-out that was bounded earlier — a node with many channels fans out one `graph/node` request per channel at once.

### Suggested fix

Mirror the existing helper: build task thunks and run them through `common.runWithConcurrencyLimit(tasks, 20, done)` (already used by the CLN peers/route and LND peers paths after #1629), guarding each `done` response-send with a try/catch as in that PR.

### Notes

- Eclair needs no change: it resolves aliases inline from a bulk nodes list, no per-peer/-channel fan-out.
- The immediate `EAGAIN` trigger is CLN-specific (clnrest), so these LND sites are a parity/robustness improvement rather than a confirmed production failure — hence a separate follow-up rather than part of #1629.

7-falseparfait:
Fixes #1630

Mirrors the `runWithConcurrencyLimit(tasks, 20, done)` pattern introduced
in #1629 across the remaining unbounded `Promise.all(map(...))`
alias-resolution fan-outs in the LND graph and channels controllers, for
parity with the CLN and LND peers paths.

Sites fixed:
- `server/controllers/lnd/graph.ts`: `getQueryRoutes`, `getAliasesForPubkeys`
- `server/controllers/lnd/channels.ts`: `getAllChannels`, `getPendingChannels`, `getClosedChannels`

Since `runWithConcurrencyLimit` resolves per-task failures into the
results array instead of rejecting the whole batch, `getQueryRoutes` and
`getAliasesForPubkeys` now guard against `{error}` entries in the
results and fall back to `'Unknown'` for that alias rather than leaking
the error object into the response.

Eclair is untouched, as noted in the issue (it resolves aliases inline
from a bulk nodes list, no per-peer/-channel fan-out).

Verified: `npm run lint`, `npm run test` (204/204 passing),
`npm run buildbackend`, and `npm run buildfrontend` all pass cleanly.


saubyk:
Heads up: I rebased this branch onto `Release-0.15.10` and force-pushed it. Apologies for rewriting your branch — the PR was unmergeable otherwise, and none of it was caused by your change.

### What happened

0.15.9 was published before this landed, so I retargeted the PR from `Release-0.15.9` to `Release-0.15.10`. That exposed a history problem:

- You branched off `Release-0.15.9`.
- `Release-0.15.9` was merged into `master` via **rebase**, which rewrote all 70 of its commits to new hashes — identical content, new identity.
- `Release-0.15.10` was cut from `master`, so it descends from those rewritten copies.

Git could no longer find a recent common ancestor between your branch and its new base, so the merge base fell back to `1a6cb0b0` — the commit just before the 0.15.9 cycle started. GitHub then rendered the entire cycle as part of this PR: **71 commits, 197 files, +5300/−5949**, in a `CONFLICTING` state. Merging it would have reset `package.json` back to `0.15.9-beta`.

### What I did

```
git rebase --onto Release-0.15.10 Release-0.15.9 fix/bound-lnd-alias-fanout-1630
```

Your commit replayed with zero conflicts. The PR is now **1 commit, 4 files, +135/−104** and `MERGEABLE`.

I compared the patch before and after the rebase (`Release-0.15.9..<old head>` vs `Release-0.15.10..<new head>`) — byte-for-byte identical, and you remain the commit author. Nothing in your change was altered.

Re-verified on the rebased branch:

- `npm run lint` — clean
- `npm run test` — 204/204 passing
- `npm run buildbackend` — clean, and the committed `backend/` artifacts match the build output exactly

### What you need to do

Your local clone is now out of sync with the branch. Before doing any further work on it:

```
git fetch origin
git reset --hard origin/fix/bound-lnd-alias-fanout-1630
```

(Use `reset --hard`, not `git pull` — a pull would merge the old history back in and reintroduce the noise. Substitute whichever remote points at your fork if it isn't `origin`.)

Thanks for the fix, and sorry for the detour.

saubyk (commented):
Another thing to note:

Missing release notes entry. Ideally for this change it should go under Code Health category

7-falseparfait:
Thanks for the thorough review, @saubyk  Pushed commits
addressing this:

F1 (shared options race) — Fixed. getAliasForChannel and
getAliasFromPubkey now take an explicit requestOptions parameter
instead of closing over the module-level options, and each handler
captures a per-request copy before building the task thunks.

F3, F4, F5 (empty-array hang, missing catch, result ordering) — These
are false positives against runWithConcurrencyLimit's actual
implementation in server/utils/common.ts (not in this PR's diff, so
understandably flagged as unverifiable):
- Empty tasks short-circuit immediately via
  `if (!tasks || tasks.length === 0) { return finish(); }`
- The helper is callback-based, never returns a promise, and catches
  every task's rejection internally into `results[i] = { error: err }`
  — nothing escapes as an unhandled rejection.
- Results are written back at their original index
  (`results[currentIndex] = result`), not pushed on settle, so
  ordering is preserved regardless of completion order.
Happy to paste the full helper source here if useful for future
reviews of code that depends on it.

F6 — Fixed. All 5 catch blocks now mirror closeChannel's hardcoded
res.status(500) pattern instead of routing raw exceptions through
handleError.

F8 — Fixed. Switched to the suggested typeof v === 'string' check for
robustness, even though currently unreachable given
getAliasFromPubkey's internal catch.

F2 — Noted, didn't change scope here; happy to open a follow-up issue
to check CLN channel-list parity if that'd be useful.

F7 — Agreed pre-existing/out of scope for this PR.

Also added a release notes entry under Code Health, and merged in
upstream/Release-0.15.10 to pick up the #1653 dependency update and
resolve a release-notes.md conflict from that PR — this branch is now
conflict-free against the base.


Re-verified after all changes: 

npm run lint (clean), npm run test
(204/204, twice in a row to rule out an earlier one-off Chrome headless
disconnect), npm run buildbackend, npm run buildfrontend.

7-falseparfait:
Addressed the remaining review feedback:

- F1: Each alias lookup task now receives its own shallow copy of
  `requestOptions`, and `selectedNode` is captured into a local constant
  before fan-out to avoid sharing mutable state across concurrent tasks.

- F9: Alias lookup requests no longer inherit the caller's query string.
  `qs` is excluded from the request options snapshot used for alias
  sub-requests.

F11 — Ran the docker/regtest fixture (built a local image from this
branch: `docker build -t rtl-local:fix-1630 . && RTL_IMAGE=rtl-local:fix-1630 docker compose up -d && ./scripts/seed.sh`) and verified against real LND nodes:

- Channels page (bob, who has real channels to alice and carol post-seed):
  channels list correctly with resolved aliases.
- Pending Channels tab (all seeded channels are confirmed/active, so this
  hits the empty-task-list path directly): loads instantly with 0 pending
  across all four categories — no hang.
- Query Routes (alice -> carol via bob): routes correctly; hop aliases
  resolve when available in the local graph and gracefully fall back to an
  empty string when they are not yet propagated in regtest gossip.
- No RTL errors observed during testing beyond expected startup timing
  noise before the LND nodes were ready.

Eclair failed to come up in this fixture run for an unrelated reason
(`curl: (52) Empty reply from server during channel-open setup`), which
didn't block verification of the LND changes.

F10: I didn't make changes for this because it concerns the CLN/LND-peers
controllers outside the scope of this PR. Happy to investigate those in a
follow-up if needed.

PR #1668, opened 2026-08-04
Fix options race in getPeers/postPeer alias resolution (lnd/peers.ts)

### LINKED ISSUE #1662: getAliasForPeers reads module-level options inside limiter-deferred tasks (lnd/peers.ts)
Follow-up from the review of #1651. **Pre-existing** since #1629, and deliberately left out of #1651's scope, which fixed the same shape only in `lnd/graph.ts` and `lnd/channels.ts`.

## What the code does

`server/controllers/lnd/peers.ts` keeps a module-level `let options = null` (`:5`) that every handler in the file reassigns wholesale with `options = common.getOptions(req)` (`:23`). `getAliasForPeers` (`:9`) does not receive it — it mutates `options.url` and hands that same shared object to `request()` at call time:

```ts
export const getAliasForPeers = (selNode: SelectedNode, peer) => {
  options.url = selNode.settings.lnServerUrl + '/v1/graph/node/' + peer.pub_key;
  return request(options).then(...)
```

Both call sites now defer those lookups across event-loop turns through the concurrency limiter — `getPeers` (`:31-32`) and `postPeer` (`:64-65`).

## Why it matters

`common.getOptions` returns a **reference** to `req.session.selectedNode.authentication.options` (`server/utils/common.ts:178-185`), not a fresh object, so the module-level `options` *is* the selected node's live auth object — macaroon header included.

Under the old `Promise.all(map(...))` every alias request was issued in one synchronous tick, so no other handler could interleave. With `runWithConcurrencyLimit` the tasks run in waves of 20 across many turns, and each task reads `options` at dispatch time. In a multi-node deployment, a second request handled by this same file mid-fan-out — the RTL UI polls peers, and any second session or node switch will do it — repoints `options` at the *other* node's auth object. Pending tasks then set `url` from their own `selNode` while sending the other node's credentials: node A's macaroon shipped to node B's endpoint. The failure is silent, because `getAliasForPeers` swallows its rejection and falls back to the pubkey prefix (`:15-18`), so the response is a normal `200` with unresolved aliases.

## Mitigating context

Both routes are behind `isAuthenticated`, and the misdirected credential goes to another node the same operator already configured — so this is a trust-boundary/robustness fix, not an external disclosure. It needs a genuine multi-node setup and concurrent traffic to trigger.

## Suggested scope

Mirror what #1651 did for graph/channels:

- Give `getAliasForPeers` an explicit `requestOptions` parameter instead of reading module state.
- Snapshot the options object per request before building the task thunks, and hand each task its own shallow copy (`{ ...requestOptions }`), so no two in-flight tasks share the object they assign `url` on.
- Drop `qs` from that snapshot, as #1651 does, so alias sub-requests don't inherit the caller's query string.

Note the residual in #1651's version worth improving here: the snapshot is taken inside the outer `.then`, which narrows the race window to the outer request's duration rather than closing it. Taking it right after `options = common.getOptions(req)` (and after the handler's own `url`/`qs` assignments) closes it outright.

## Not affected

- **CLN** — `cln/network.ts:95` already builds a self-contained request from `selNode.authentication?.options`, with a comment explaining exactly this hazard.
- **Eclair** — resolves aliases inline from a bulk nodes list, no per-peer fan-out.
- **LND graph/channels** — fixed in #1651.

7-falseparfait:
fixes #1662

follow-up to #1651. threads an explicit requestOptions parameter
through getAliasForPeers in server/controllers/lnd/peers.ts instead of
reading the module-level options, closing the same shared-mutable-state
race #1651 fixed in graph.ts/channels.ts.

what changed

getAliasForPeers now takes an explicit requestOptions parameter and
uses it for the request, instead of mutating and reading the
module-level options.

getPeers and postPeer now take the requestOptions snapshot
synchronously, right after their own url/form assignments and before
the first request()/request.post() call, not inside the outer .then()
like #1651 did for graph/channels. everything up to that point runs in
one synchronous tick, so no concurrent request to a different node can
reassign the shared options object before the snapshot happens. this
closes the race window fully instead of just narrowing it.

each task thunk gets its own shallow copy ({ ...requestOptions }), and
selNode is captured locally instead of read lazily inside each thunk,
same as #1651's F1 follow-up.

deletePeer is untouched, single request, no fan-out, nothing to fix
there.

qs isn't set anywhere in peers.ts so the qs-exclusion destructure is a
no-op here, kept for parity with #1651's files.

verification

npm run lint, npm run test (twice: backend 38/38, frontend 204/204
both times), npm run buildbackend, npm run buildfrontend, all clean.

also ran against the docker/ regtest fixture (built a local image from
this branch, RTL_IMAGE=rtl-local:fix-1662 docker compose up -d then
seed.sh). with alice and bob both live, repeatedly navigated between
their peers pages in two tabs to approximate concurrent multi-node
access, each tab consistently showed its own correct peers/aliases the
whole time, no errors in rtl's logs. this is a manual approximation not
a guaranteed race reproduction, the real correctness evidence is the
synchronous snapshot placement itself, verified against the exact
ordering in the diff.

### DIFF
--- a/backend/controllers/lnd/peers.js
+++ b/backend/controllers/lnd/peers.js
@@ -4,17 +4,14 @@ import { Common } from '../../utils/common.js';
 let options = null;
 const logger = Logger;
 const common = Common;
-export const getAliasForPeers = (selNode, peer, requestOptions) => {
-    requestOptions.url = selNode.settings.lnServerUrl + '/v1/graph/node/' + peer.pub_key;
-    return request(requestOptions).then((aliasBody) => {
-        logger.log({ selectedNode: selNode, level: 'DEBUG', fileName: 'Peers', msg: 'Alias Received', data: aliasBody.node.alias });
-        peer.alias = aliasBody.node.alias;
-        return aliasBody.node.alias;
-    }).catch((err) => {
-        peer.alias = peer.pub_key.slice(0, 20);
-        return peer.pub_key;
-    });
-};
+export const getAliasForPeers = (selNode, peer, requestOptions) => request({ ...requestOptions, url: selNode.settings.lnServerUrl + '/v1/graph/node/' + peer.pub_key }).then((aliasBody) => {
+    logger.log({ selectedNode: selNode, level: 'DEBUG', fileName: 'Peers', msg: 'Alias Received', data: aliasBody.node.alias });
+    peer.alias = aliasBody.node.alias;
+    return aliasBody.node.alias;
+}).catch((err) => {
+    peer.alias = peer.pub_key.slice(0, 20);
+    return peer.pub_key;
+});
 export const getPeers = (req, res, next) => {
     logger.log({ selectedNode: req.session.selectedNode, level: 'INFO', fileName: 'Peers', msg: 'Getting Peers..' });
     options = common.getOptions(req);
@@ -23,8 +20,8 @@ export const getPeers = (req, res, next) => {
     }
     options.url = req.session.selectedNode.settings.lnServerUrl + '/v1/peers';
     const selNode = req.session.selectedNode;
-    const { qs: _qs, ...requestOptions } = options;
-    request(options).then((body) => {
+    const { qs: _qs, form: _form, ...requestOptions } = options;
+    request({ ...requestOptions }).then((body) => {
         logger.log({ selectedNode: req.session.selectedNode, level: 'DEBUG', fileName: 'Peers', msg: 'Peers List Received', data: body });
         const peers = !body.peers ? [] : body.peers;
         // Bound concurrent alias lookups so a node with many peers can't fire one graph/node
@@ -61,11 +58,10 @@ export const postPeer = (req, res, next) => {
         perm: perm
     });
     const selNode = req.session.selectedNode;
-    const { qs: _qs, ...requestOptions } = options;
-    request.post(options).then((body) => {
+    const { qs: _qs, form: _form, ...requestOptions } = options;
+    request.post({ ...requestOptions, form: options.form }).then((body) => {
         logger.log({ selectedNode: req.session.selectedNode, level: 'DEBUG', fileName: 'Peers', msg: 'Peer Connected', data: body });
-        options.url = req.session.selectedNode.settings.lnServerUrl + '/v1/peers';
-        request(options).then((body) => {
+        request({ ...requestOptions, url: selNode.settings.lnServerUrl + '/v1/peers' }).then((body) => {
             const peers = (!body.peers) ? [] : body.peers;
             // Bound concurrent alias lookups (parity with the CLN fix, #1501).
             const getPeerAliasesTasks = peers.map((peer) => () => getAliasForPeers(selNode, peer, { ...requestOptions }));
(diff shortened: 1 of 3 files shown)


saubyk:
Rebased this onto `Release-0.15.11` (0.15.10 has shipped) and retargeted the PR — the diff is unchanged. If you have the branch checked out locally, please `git fetch && git reset --hard origin/fix/peers-options-race-1662` rather than `git pull`, which would re-merge the old history.

saubyk:
> Rebased this onto `Release-0.15.11` (0.15.10 has shipped) and retargeted the PR — the diff is unchanged. If you have the branch checked out locally, please `git fetch && git reset --hard origin/fix/peers-options-race-1662` rather than `git pull`, which would re-merge the old history.

One this is done, will start the round of bot review

saubyk:
One more rebase, sorry — `Release-0.15.11` itself was rebased onto `master` to pick up a docs commit, so I replayed your commit onto the new branch tip (`3559f4b0` → `b25a166a`, diff still unchanged). Same drill locally: `git fetch && git reset --hard origin/fix/peers-options-race-1662` rather than `git pull`.

7-falseparfait:
Synced locally with b25a166a, thanks @saubyk

7-falseparfait:
Pushed updates in commit `27f4f8a` addressing all findings from rtlreviewbot (F1–F6):

F6 (Major) & F3: Removed module-level options.url mutation in postPeer, updated postPeer follow-up list request and getPeers initial request to use isolated requestOptions.

F5: Stripped form payload from requestOptions for GET requests while preserving form: options.form on the POST connect call.

F1: Made getAliasForPeers caller-safe by deriving request options inline instead of mutating the argument.

Verified with npm run buildbackend, npm run lint, and npm run test (38/38 backend, 204/204 frontend). Ready for re-review @saubyk

7-falseparfait:
Addressed F1 and F7 in `server/controllers/lnd/peers.ts`:
- **F7:** Changed `options.form` to `_form` on line 59 to eliminate the remaining module-level options read.
- **F1:** Added explicit type annotations to `getAliasForPeers`.

Re-ran `npm run buildbackend`, `npm run lint`, and `npm run test` . Ready for final approval @saubyk thank youuuu

saubyk:
Reviewed `8f181bb3` — the fix is correct and complete: no read of the module-level `options` survives an async boundary in either handler, F1 and F7 are addressed, and on F8 I checked `common.getOptions`: it mutates only top-level fields per call and `headers` is built once per node, so the shallow snapshot is sufficient. `backend/` matches a fresh `tsc` build; lint clean.

To close F4 I pushed one commit on top of yours, `a43c9550`, adding `test/backend/lnd-peers.test.mjs` — two `node:test` cases (one per handler) using two fake LND HTTP servers that reject a wrong macaroon, with node A's response held while node B's request reassigns `options`. Both pass here and both fail against the pre-fix controller exactly as #1662 predicts (aliases degrade to raw pubkeys / `postPeer` 401s). Backend suite is now 40/40.

Nothing further needed from you — thanks for the quick turnaround on the review rounds. If you have the branch checked out, `git pull` is fine this time (it's a plain fast-forward).

PR #1673, opened 2026-08-10
Mark the 2FA token input as a one-time code

### LINKED ISSUE #1674: 2FA login: the token input has no autocomplete attribute, so password managers do not offer the one-time code
**Is your feature request related to a problem? Please describe.**

When two-factor authentication is on, you must type the one-time code by hand at each login. No password manager offers the code.

The cause is the token input in `src/app/shared/components/data-modal/login-2fa-token/login-2fa-token.component.html`:

```html
<input autoFocus matInput type="text" id="token" name="token" tabindex="2" required [(ngModel)]="token">
```

The input has no `autocomplete` attribute. Its `id`, `name` and label are all `token`. A password manager has no data to find the field.

Bitwarden is an example. Its `isTotpField()` function accepts a field in two conditions only:

- The field has `autocomplete="one-time-code"`.
- The `id`, the `name`, the placeholder or the label contains a word from `AutoFillConstants.TotpFieldNames`. That list has `totp`, `twofa`, `otpcode` and `verificationcode`, but not `token`.

The token input has neither property, so Bitwarden shows no suggestion.

There is a second cause. RTL makes the token dialog only after you send the password. So the field is not in the page during the first autofill operation.

**Describe the solution you'd like**

Add the standard attribute to the input:

```html
<input autoFocus matInput type="text" id="token" name="token" autocomplete="one-time-code" tabindex="2" required [(ngModel)]="token">
```

`one-time-code` is the standard value for this field. Password managers read the attribute before they use their word lists. Browsers use the same value for their own one-time code function.

You do not have to change the `id` or the `name`. No other function of the dialog changes.
**Describe alternatives you've considered**

- Change the `id` and the `name` to `totp`. Each password manager has a different word list, so this is less reliable than the standard attribute.
- Move the token field into the login form. Then the field is in the page during the first autofill operation, but this changes the two-factor flow too much.
- Make no change. Each user with two-factor authentication continues to copy each code by hand.

**Additional context**

Found in RTL 0.15.10 with the Bitwarden browser extension.

PR #1673 makes this change. It also adds a test and the new `frontend/` bundle.

lorenzolfm:
## Problem

The token input in the two-factor login dialog has no `autocomplete` attribute:

```html
<input autoFocus matInput type="text" id="token" name="token" tabindex="2" required [(ngModel)]="token">
```

Its `id`, `name` and label are all `token`. A password manager has no data to find the field.

Bitwarden is an example. Its `isTotpField()` function accepts a field in two conditions only:

- The field has `autocomplete="one-time-code"`.
- The `id`, the `name`, the placeholder or the label contains a word from `AutoFillConstants.TotpFieldNames` (`totp`, `twofa`, `otpcode`, `verificationcode`).

The token input has neither property, so Bitwarden shows no suggestion.

RTL also makes this dialog only after you send the password. So the field is not in the page during the first autofill operation. Users type each code by hand.

## Change

Add `autocomplete="one-time-code"` to the input. This is the standard value for this field. Password managers read the attribute before they use their word lists, and browsers use the same value for their own one-time code function.

You do not have to change the `id` or the `name`. The change is in the frontend only, and the token check does not change. The dialog is common to LND, Core Lightning and Eclair.

## Tests

- `npm run lint` is clean.
- All 205 tests pass with `npm run test`. There were 204 tests before this change.
- The new test is a true guard. If you remove the attribute, the test fails (`1 FAILED, 204 SUCCESS`).
- I built `frontend/` with `npm run buildfrontend` after `npm ci --legacy-peer-deps`. The build is repeatable: each lazy chunk (`853.`, `190.`, `17.`, `193.`) and `styles.` kept its hash. Only `main.*`, `runtime.*` and two script tags in `index.html` are different. The attribute is in the new bundle.

## Base branch

`CONTRIBUTING.md` asks for the current `Release-x.y.z` branch. But version 0.15.10 is complete and merged into `master`, and 0.15.11 has no branch yet. So I used `master`. Tell me if you want a different branch, and I will rebase the commit.

The release note is in `release-notes/Release-notes-0.15.11.md`. Tell me if you want it in a different file.

Fixes #1674.


saubyk:
Thanks — this is a clean, well-argued change. Two maintainer-side adjustments, no action needed from you on the content:

- Rebased onto `Release-0.15.11` (PRs flow through the release branch, not `master`) and retargeted. The release-note file already existed there from #1676, so your `## Enhancements` entry now sits above the existing `## Code Health` one; your `src/` and spec changes are unchanged.
- Replaced the `frontend/` bundle with one regenerated from your sources on the release branch's toolchain (added as a separate commit on top of yours). That's standing policy for external PRs — committed build output can't be reviewed by reading the diff, so the maintainer rebuilds it. For the record, a token-level diff of your bundle against `master`'s showed exactly the two added strings, so yours was fine.

If you have the branch checked out locally, please `git fetch && git reset --hard origin/fix/2fa-token-autocomplete` rather than `git pull`.

lorenzolfm:
Thanks @saubyk

Issue #1220, opened 2023-04-18
Error 490 running "node rtl"


Foxtrot-Zulu:
After installing and running c-lightning-REST-0.10.2, I installed the RTL but when I tried to run it, I got this error:

**Screenshots**
![node_rtl](https://user-images.githubusercontent.com/111104979/232648411-848e53e6-e34e-4619-9670-b97ff25330bb.png)

After going through the logs/error msgs, I found that the module "request-promise" can't be located and it is no longer available on github.

Any ideas?

**my environment**
* Version of `RTL` = 0.13.6
* Version of `core lightning` = 23.02.2
* Version of `bitcoind` = 24.0.1
* Version of `node.js` = 19.9.0
* Operating system (`uname -a` on *Nix) = Ubuntu 22.04 Linux 5.15.0-69-generic #76-Ubuntu SMP Fri Mar 17 17:19:29 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

Thanks in advance.

Foxtrot-Zulu:
Thanks for the prompt response. I did npm install --legacy-peer-deps

ShahanaFarooqui:
It doesn't make sense then. Could you please try installing `request-promise` explicitly with `npm install request-promise --legacy-peer-deps` command?

If that doesn't work, the only other possible solution to try is to delete node_modules and try a fresh install.

Foxtrot-Zulu:
I did the fresh install of node.js (18.16.0) npm (9.5.1) and then RTL.  This is the result:

![node_rtl2](https://user-images.githubusercontent.com/111104979/232897443-6f07792c-afa1-4b43-a7e8-ec7a72264497.png)

Now when I do: "npm rtl" this is the outcome:

![node_rtl3](https://user-images.githubusercontent.com/111104979/232898053-c8ca1f8a-a20c-450a-bd40-b568c4575da6.png)

ShahanaFarooqui:
Can you verify the path and check that rtl.js file is in this location?

Foxtrot-Zulu:
it is not.

ShahanaFarooqui:
> 

Then that is the issue. RTL is not installed properly on this location. Otherwise rtl.js file will exist in the parent folder.

Foxtrot-Zulu:
[...]

PR #1681, opened 2026-08-20
Declare a minimum Node.js version

saubyk:
`package.json` carried no `engines` field, so npm installed RTL on any Node.js version without a word, and an unsupported runtime only surfaced later as a confusing failure at `node rtl` — as in #1220, where two users on Node 19.x/20.0.0 got `Cannot find module 'request-promise'` from an install that had reported success, and both were fixed by dropping to Node 18.15. (The dependency behind that particular report is gone: #1638 replaced `request`/`request-promise` with axios in 0.15.9.)

This sets `engines.node` to `>=20.19.0`, the floor the Angular 20 toolchain requires, so npm prints an `EBADENGINE` warning at install time instead. The docs disagreed with each other as well — `CONTRIBUTING.md` said "v14 & above" and the README gave no version at all; both now state the same floor.

### Judgment call

I used the plain `>=20.19.0` floor rather than Angular's exact `^20.19.0 || ^22.12.0 || >=24.0.0`, so odd-numbered Node lines (21, 23) don't emit a warning on every install. It still rejects the versions that broke #1220. This is advisory only — no `.npmrc` with `engine-strict`, so npm warns and continues rather than blocking anyone on a working-but-unusual Node.

### Verification

- `npm run lint` — clean
- backend tests — 40/40 pass
- frontend specs — 205/205 pass

No `frontend/` or `backend/` change: nothing under `server/` or `src/` was touched and the app version is unchanged, so neither artifact moves.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -45,7 +45,7 @@ Docker Image: https://hub.docker.com/r/shahanafarooqui/rtl
 
 ### <a name="prereq"></a>Prerequisites
 * Functioning and synced LND lightning node.
-* Node.js, which can be downloaded [here](https://nodejs.org/en/download/)
+* Node.js v20.19 or above, which can be downloaded [here](https://nodejs.org/en/download/)
 * Recommended Browsers: Chrome, Firefox, MS Edge
 
 ### <a name="install"></a>Installation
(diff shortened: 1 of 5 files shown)


ShahanaFarooqui (approved):
LGTM!!!

Issue #1597, opened 2026-05-12
Multi-node RTL-Config.json gets overwritten when updating application settings


CosimoRicciardi:
# RTL issue: multi-node configuration gets overwritten when updating application settings

## Environment

* RTL latest master
* Node.js v22
* Multi-node setup:

  * LND node
  * CLN node using rune authentication
* RTL running via systemd on Linux

## Problem

In a multi-node configuration, RTL overwrites `RTL-Config.json` and silently removes nodes when application settings are updated from the UI.

Example:

Initial config:

```json
"nodes": [
  {
    "index": 1,
    "lnNode": "[alias-lnd]",
    "lnImplementation": "LND"
  },
  {
    "index": 2,
    "lnNode": "[alias-cln]",
    "lnImplementation": "CLN"
  }
]
```

After opening Settings or changing application settings, the config gets rewritten containing only the currently selected node.

Depending on which node is active, the other node disappears from the config and from the UI.

This makes RTL unstable for mixed LND + CLN environments because the configuration file becomes progressively corrupted during normal UI usage.

---

# Root cause

The issue appears to originate from `updateApplicationSettings()` inside:

```text
/backend/controllers/shared/RTLConf.js
```

RTL rebuilds the configuration object from the frontend payload:

```js
const config = common.addSecureData(req.body);
common.appConfig = JSON.parse(JSON.stringify(config));
```

The frontend payload only contains the currently selected node, so all other nodes are discarded when the config is rewritten.

Later:

```js
fs.writeFileSync(RTLConfFile, JSON.stringify(config, null, 2), 'utf-8');
```

persists the truncated configuration to disk.

---

# Patch applied

## 1. Preserve existing node list

Patched `updateApplicationSettings()` to merge the incoming config with the previous config instead of replacing it completely.

Added:

```js
const oldConfig = JSON.parse(fs.readFileSync(RTLConfFile, 'utf-8'));
const config = common.addSecureData(req.body);

if (oldConfig.nodes && oldConfig.nodes.length > 0 && config.nodes && config.nodes.length > 0) {
    oldConfig.nodes.forEach((oldNode) => {
        if (!config.nodes.find((newNode) => newNode.index === oldNode.index)) {
            config.nodes.push(oldNode);
        }
    });
}
```

Then replaced:

```js
common.appConfig = JSON.parse(JSON.stringify(config));
```

with:

```js
const mergedConfig = JSON.parse(JSON.stringify(oldConfig));

[...]

PR #1682, opened 2026-08-20
Open an LND channel with the entire wallet balance

### LINKED ISSUE #155: Should be able to open a channel and sweep entire wallet into it - taking fee into account
**Is your feature request related to a problem? Please describe.**
Should be able to completely empty your balance into a channel bringing your wallet balance to 0. 

**Describe the solution you'd like**
Add a check box to "sweep entire balance into channel".

**Describe alternatives you've considered**
show in interface the total amount with fee that will need to open channel.

saubyk:
Fixes #155.

Emptying the on-chain wallet into a channel could not be done from the UI. The amount had to be typed in, and the balance itself is never a valid amount — nothing would be left for the on-chain fee, or for the reserve LND holds back for anchor channels. Only the node can work the maximum out, which is why this issue waited on [lnd#6903](https://github.com/lightningnetwork/lnd/pull/6903) (`fund_max` on `OpenChannelRequest`, LND 0.16.0) rather than being fixed with UI arithmetic.

### What changed

- **`server/controllers/lnd/channels.ts`** — `postChannel` accepts `fund_max` and sends either it or `local_funding_amount`, never both; LND rejects a request carrying the two together.
- **Open-channel modal and connect-peer stepper** — both LND channel-open paths get a "Use Entire Wallet Balance" toggle that disables and clears the amount field, and relaxes its `required`/`max` validation while on.
- **Compatibility** — the flag is only sent when the toggle is on, and the toggle only renders on LND 0.16.0 and above (same `isVersionCompatible` gate the Taproot toggle uses). Nothing changes for older nodes.

### Other implementations

Core Lightning already has the equivalent — its dialog sends `amount: "all"`. Eclair's `/open` takes an explicit `fundingSatoshis` with no fund-max option, so it is unchanged.

### Tests

- `test/backend/lnd-channels.test.mjs` — two tests against a fake LND that record the body actually sent, pinning which of `fund_max` / `local_funding_amount` reaches the node. Confirmed the first fails against the pre-change controller.
- Component specs for the version gate and the dispatched payload in both dialogs. Mutation-checked: forcing `fundMax: false` into the payload fails the fund-max spec and nothing else.

### Verification

- `npm run lint` — clean
- backend tests — 42/42 pass
- frontend specs — 213/213 pass (205 before, 8 new)
- `frontend/` and `backend/` regenerated from these sources; only the LND chunk, `main`, `runtime` and `index.html` move.

Not yet exercised against a live node — I'll run the regtest fixture before this merges.


saubyk:
Verified against the `docker/` regtest fixture (LND 0.20.0-beta), running this branch's image and driving RTL's own API.

**Fund max, through RTL.** alice's wallet before: `total_balance` 4,745,574 sats, `reserved_balance_anchor_chan` 30,000. `POST /api/lnd/channels` with `fund_max: true` and no amount → **201**. After confirmation:

```
alice wallet:  total_balance 40000, confirmed 40000, locked 0, reserved_balance_anchor_chan 40000
alice channel: capacity 4705266, local_balance 4701796, initiator true
```

4,745,574 − 4,705,266 = 40,308 — the 40,000 anchor reserve the node holds back (it rises as anchor channels are added) plus ~308 sats of on-chain fee at 2 sat/vByte. Spendable balance is now zero, which is what #155 asked for, and the split between fee and reserve is exactly the arithmetic RTL cannot do for itself.

**Why the either/or matters.** Straight to LND, bypassing RTL, with both fields set:

```json
{"code": 2, "message": "cannot use a local funding amount and fundmax parameters"}
```

That is the request `postChannel` is written never to send.

**No regression on the ordinary path.** An explicit `local_funding_amount: 250000` open through the same API → 201, channel confirmed at capacity 250,000.

**One wrinkle worth knowing about, not a regression.** Asking for fund max with nothing spendable produces a confusing message straight from LND — `not enough witness outputs to create funding transaction, need -0.00040000 BTC only have 0 BTC available` (note the negative). It surfaces in the dialog's error area like any other open failure. Disabling the toggle when the wallet has no spendable balance would read better; happy to add that here or leave it.

saubyk:
Added in 258a9bc6 — the toggle is now disabled while nothing is spendable.

The guard uses `total_balance` less `reserved_balance_anchor_chan` from the blockchain balance in the store, rather than `total_balance` alone. That distinction is the whole point: the case that produced the negative-amount error was a wallet reporting 40,000 sats where every one of them was the anchor reserve. A plain `total_balance <= 0` check would have missed it. `reserved_balance_anchor_chan` was added to the `BlockchainBalance` model — LND already returns it and RTL passes the response through untouched.

Both dialogs pick it up, with a "No spendable balance" note beside the disabled toggle. The amount field's existing `[max]="totalBalance"` is deliberately left alone — it has the same reserve blind spot, but changing it is a separate question from this PR.

Specs cover the derivation (a wallet that is all reserve reads as zero spendable) and the disabled state in both dialogs; mutation-checked by removing the `[disabled]` binding, which fails that spec and nothing else.

`npm run lint` clean, backend 42/42, frontend 216/216 (213 before, 3 new). Artifacts rebuilt. Not re-run against the fixture — this is UI-only and the node-side behaviour is unchanged from the run reported above.

saubyk:
Folded the two fixture gotchas into this PR in ec5e8760, since both surfaced while verifying this change against regtest. They go into `docker/README.md` and the `rtl-docker-fixture` skill, which had neither.

**Mining needs `-rpcwallet=rtldev`.** The eclair wallet means two wallets are loaded, so a bare `bin/b-cli -generate 6` fails with `error code: -19  Multiple wallets are loaded`. Piping the call takes the error with it, so it reads as a successful mine that produced no blocks — the symptom only shows up later as a channel that will not confirm. That is exactly how it caught me: several minutes of wondering why a fund max channel stayed pending.

**An idle network reports itself out of sync.** LND judges sync from the age of the chain tip, not its height, so a fixture left running overnight shows a red "Not Synced to Chain" in RTL while sitting at precisely bitcoind's height. One block clears it. Nothing to debug, and it never happens on a real chain.

Docs only — no release-note entry, matching how the fixture has been treated before. Say the word if you would rather these went to `master` separately.

saubyk:
Correction to my verification note above: the node switch was **not** failing. I called `/rtl/api/rtlconf/updateSelNode/...`, but the route is mounted at `/conf` (`server/routes/shared/index.ts:15`), so the correct path is `/rtl/api/conf/updateSelNode/<idx>/<prev>`. A wrong API prefix falls through to the SPA catch-all and returns index.html with a 200, which is why it looked like a successful switch that did nothing.

So there is nothing here for anyone to look at, and no connection to #1597 — please disregard that suggestion. The verification results themselves are unaffected: every call in them went to the default node, which is what the numbers describe.

saubyk:
Closing out this round. **F22** is fixed in 42acb1f4; **F20** and **F19** are answered by testing against the regtest fixture rather than by changing code, and **F21** is declined. Details below.

### F22 — fixed

The modal's two gate specs now assert the DOM in both directions, matching the connect-peer specs. Mutation-checked: replacing the row's `*ngIf` condition with `true` now fails the below-0.16.0 case, which it did not before.

### F20 — verified, no change

> One check against the real `/lnd/balance/blockchain` payload on the regtest fixture settles it.

Ran exactly that, against alice on the `docker/` fixture (LND 0.20.0-beta), through RTL's own controller:

```json
{"total_balance":"4995882","confirmed_balance":"4995882","unconfirmed_balance":"0",
 "locked_balance":"0","reserved_balance_anchor_chan":"20000",
 "account_balance":{"default":{"confirmed_balance":"4995882","unconfirmed_balance":"0"}}}
```

All three field names `getSpendableBalance` depends on survive verbatim, with a non-zero `reserved_balance_anchor_chan` on a node that has anchor channels. Both silent failure modes are ruled out.

The same fixture run also exercised the guard end to end. With a deposit sent to carol and deliberately left unconfirmed:

```
balance: total 2,020,000 | confirmed 0 | unconfirmed 2,020,000 | reserved 20,000
before this PR (total - reserve)     :  2,000,000  => toggle enabled
after  this PR (confirmed - reserve) :    -20,000  => toggle disabled
```

and sending `fund_max` in that state returns, from LND:

```
not enough witness outputs to create funding transaction,
need -0.00030000 BTC only have 0 BTC available
```

which is the negative-amount error the gate exists to prevent.

### F19 — declined; its open question is answered

> Whether LND's grpc-gateway coerces that or rejects it is not determinable from this repo […] so it is either an announced-vs-unannounced channel opened against the caller's intent or an opaque upstream error.

It is the second, and it is determinable — a urlencoded `POST /api/lnd/channels` against the fixture returns:

```
proto: (line 1:102): invalid value for bool type: "false"
```

Offset 102 lands on `private`, not `fund_max`: the fund-max branch resolved correctly and sent `local_funding_amount` with no `fund_max`. So the request fails loudly at the node rather than opening a channel against the caller's intent — no silent misbehaviour.

It is also pre-existing: `private: privateChannel` is on line 145 of `Release-0.15.12`, unchanged by this PR, and it breaks *any* urlencoded caller regardless of `fund_max`. RTL's own frontend sends JSON booleans, so nothing user-facing is affected. Out of scope here; better as its own issue covering all three fields together.

### F21 — declined

Confirmed the mechanism: `catchError` dispatches `VOID` and leaves the slice untouched, so a failed `FetchBalance` does leave `spendableBalance` at 0 while `totalBalance` is non-zero. The direction is fail-closed, the user can still open a channel by typing an amount, and the dialog is normally opened with a balance already fetched. Not worth widening the diff for at this stage of the review cycle.

saubyk:
/rtl dismiss F20 per https://github.com/Ride-The-Lightning/RTL/pull/1682#issuecomment-5369686031

saubyk:
/rtl dimiss F19 per https://github.com/Ride-The-Lightning/RTL/pull/1682#issuecomment-5369686031

saubyk:
/rtl dismiss F21 per https://github.com/Ride-The-Lightning/RTL/pull/1682#issuecomment-5369686031

PR #1683, opened 2026-08-20
fix(settings): allowlist payload fields and pin credential paths (#1660)

### LINKED ISSUE #1660: updateApplicationSettings trusts request body for node auth paths and lnServerUrl
Follow-up from the review on #1659 (finding F2). **Pre-existing** trust-boundary issue in `server/controllers/shared/RTLConf.ts`, deliberately kept out of that PR's scope.

## What the code does

`updateApplicationSettings` merges request-supplied nodes wholesale:

- `authentication: { ...oldNode.authentication, ...newNode.authentication }` — every client-sent `authentication.*` field is accepted (`macaroonPath`, `runePath`, `lnApiPassword`), as is `settings.lnServerUrl`.
- `addSecureData` re-pins server-held paths only for node indices already present in `common.appConfig.nodes`, and only for a subset of fields — so a new index, or an existing index sent with a mismatched JSON type (`"1"` vs `1`, since the map lookup is SameValueZero), is entirely caller-defined.

## Why it matters

The sinks in `server/utils/common.ts` use those caller-controlled values with the server's authority: `setOptions()` and `getAllNodeAllChannelBackup()` read `macaroonPath`/`runePath` from disk and attach the contents as auth headers on requests to `settings.lnServerUrl`. An authenticated caller can therefore point a node entry at an arbitrary readable file path and an arbitrary host, and have the server ship the file contents to that host on subsequent node requests (confused deputy).

Relatedly, `secret2FA` is accepted from the request body with no server-side pinning (the restore guard in `addSecureData` only re-assigns when the values already match), so a session holder can set the TOTP seed to a value they control. (The echo-wipe variant of this was fixed in #1659.)

## Mitigating context

The route is behind `isAuthenticated` (`server/routes/shared/RTLConf.ts:15`), so the caller already holds a full-privilege session — this is hardening of a trust boundary, not an unauthenticated exploit. Contrast the sibling `updateNodeSettings`, which allowlists only `boltzMacaroonPath`/`swapMacaroonPath` from `req.body.authentication`.

## Suggested scope

- Allowlist the `authentication.*` and `settings.*` fields accepted from the request body, mirroring `updateNodeSettings`.
- Normalize node index types when matching request nodes to server nodes.
- Validate `lnServerUrl` shape before persisting.

7-falseparfait:
## Here is a full breakdown

Fixes #1660

## Context & Problem Statement

Prior to this fix, the `/api/rtlconf/updateApplicationSettings` endpoint merged the
request body's per-node `authentication` and `settings` into the server's runtime config
(`Common.appConfig`) without allowlisting. An authenticated caller could supply a
`macaroonPath`, `runePath`, `lnApiPassword`, `configPath` or `lnServerUrl` and re-point
credential reads (`getConfig`, `getFile`) or credentialed outbound requests at
attacker-controlled values — a confused-deputy vector into host file reads and internal
HTTP targets.

## Solution Overview

The endpoint now allowlists the node fields it accepts from the body and re-pins server
URLs and credential paths to the server-held values for existing nodes:

* **Payload allowlisting:** per-node `settings` keys are restricted to the same set the
  Node Settings endpoint uses, and `authentication` is stripped wholesale from the
  request body (addSecureData re-pins every credential anchor from the server-held node).
  Unknown top-level keys, unknown node indexes (there is no add-node flow through this
  endpoint), a caller-invented node top-level key, and a plaintext `multiPass` are
  dropped with a warning instead of being persisted.
* **Credential & path pinning:** `macaroonPath`, `runePath`, `lnApiPassword`,
  `configPath`, `bitcoindConfigPath`, `channelBackupPath`, `logFile` and the server URLs
  (`lnServerUrl`, `swapServerUrl`, `boltzServerUrl`) are pinned to the server-held values
  for existing nodes and stripped from unknown nodes. Known-vs-unknown and the rebuild
  both resolve against the single authoritative runtime node list (`common.nodes`), so
  the two layers can never disagree about which node is which.
* **URL format validation:** user-editable URL settings — `blockExplorerUrl` here, plus
  `swapServerUrl`/`boltzServerUrl` on the Node Settings endpoint — are validated by the
  shared `filterNodeSettings`/`isValidHttpUrl` helpers; malformed or non-http(s) values
  (e.g. `file://`) are silently dropped and the previous server-held value is kept. This
  is format validation only and does not stop a caller from pointing at an internal host:
  self-hosted block explorers and Loop/Boltz servers are intended features, so the fields
  are deliberately not pinned.
* **Loop/Boltz service settings stay editable:** the Node Settings endpoint (the
  node-config Services page) still accepts `swapServerUrl`, `boltzServerUrl` and
  `swapMacaroonPath`/`boltzMacaroonPath`, now with the same format validation; an empty
  macaroon path clears the field. The LN credential paths have no writable path through
  any endpoint.
* **Index & path hardening:** node indexes are normalized to numbers, and values that a
  naive coercion would turn into `0` (`''`, `false`, `[]`, `null`) are treated as unknown
  instead of merging into a real node; the `defaultNodeIndex`/`selectedNodeIndex` scalars
  are held to the same known-index standard as the node entries, so an out-of-range or
  unparseable value is dropped rather than persisted across a restart; CLN's live rune
  (`authentication.runeValue`) is pinned for known nodes and stripped for unknown nodes
  like every other credential; the `getFile` channel-branch file name sanitizes
  the channel value so it cannot walk out of the backup directory; a live 2FA seed cannot
  be overwritten by a request-supplied seed.

## Key Technical Changes

* `server/controllers/shared/RTLConf.ts` — allowlists, shared
  `filterNodeSettings`/`isValidHttpUrl` helpers, `common.nodes`-based known-vs-unknown
  filtering and rebuild, channel-backup file-name sanitization in `getFile`.
* `server/utils/common.ts` — `addSecureData` pins anchors from the single authoritative
  runtime node list, never keeps a caller-supplied `multiPass`, and normalizes indexes
  without coercing empty/falsy values to `0`.

## Design Rationale

* **Defense in depth:** allowlisting plus pinning means even future settings keys cannot
  be persisted into the config file and re-parsed at the next boot/save.
* **Non-disruptive UX:** legitimate dashboard customization (themes, block explorer,
  currency, Loop/Boltz service config) keeps working; only credential paths and server
  URLs are restricted.
* **Containment:** the `getConfig` and `getFile` file reads stay confined to their
  configured roots, and the channel-branch file name can no longer escape the backup
  directory.
* **No silent no-op:** when the live runtime node list is empty the allowlisted-and-pinned
  payload is still written to the file — the save the caller is told about is the save
  that happened.

## Testing & Verification

* Backend regression tests (`test/backend/rtlconf.test.mjs`, `test/backend/common.test.mjs`)
  cover the allowlists, node-settings service edits, URL format validation (including a
  parseable `file://` URL), unknown-node drop, index normalization (`''`/`false`/`[]`/
  `null`) and the index scalars, `multiPass` rejection, `runeValue` pinning/stripping,
  runtime-node-list pinning, the empty-runtime-list path and the channel-branch
  traversal — 80 backend tests pass.
* Frontend Karma suite passes (229/229) and `npm run lint` is green.
* **Note:** invalid URL values are silently dropped rather than rejected with `400`; the
  settings-save response carries no per-field error and the UI reflects the server-held
  values on the next load.

### DIFF
--- a/backend/controllers/shared/RTLConf.js
+++ b/backend/controllers/shared/RTLConf.js
@@ -298,6 +298,22 @@ export const updateApplicationSettings = (req, res, next) => {
             }
             if (filteredNode && filteredNode.settings && typeof filteredNode.settings === 'object') {
                 filteredNode.settings = Object.fromEntries(Object.entries(filteredNode.settings).filter(([key]) => NODE_SETTINGS_ALLOWLIST.includes(key)));
+                // Validate blockExplorerUrl to reject obviously malformed values. This is format
+                // validation only — it prevents non-URL strings and non-HTTP schemes from being
+                // persisted, but does not prevent SSRF (pointing at internal IPs). A user-chosen
+                // block explorer is an intended RTL feature (self-hosted mempool instances), so
+                // this field is not pinned to the server-held value.
+                if (filteredNode.settings.blockExplorerUrl !== undefined) {
+                    try {
+                        const parsed = new URL(filteredNode.settings.blockExplorerUrl);
+                        if (!['http:', 'https:'].includes(parsed.protocol)) {
+                            delete filteredNode.settings.blockExplorerUrl;
+                        }
+                    }
+                    catch {
+                        delete filteredNode.settings.blockExplorerUrl;
+                    }
+                }
             }
             return filteredNode;
         });
--- a/backend/controllers/shared/RTLConf.js
+++ b/backend/controllers/shared/RTLConf.js
@@ -14,6 +14,29 @@ const logger = Logger;
 const common = Common;
 const wsServer = WSServer;
 const databaseService = Database;
+// The settings API echoes the whole config back, so these are the only per-node fields
+// accepted from the request body. Credential paths and runtime-only fields are not here;
+// addSecureData re-pins the credential paths for existing nodes and strips them for new
+// ones. Anything else arriving under authentication/settings is discarded.
+const NODE_SETTINGS_ALLOWLIST = [
+    'blockExplorerUrl', 'logLevel', 'logFile', 'userPersona', 'themeMode', 'themeColor',
+    'unannouncedChannels', 'fiatConversion', 'currencyUnit', 'enableOffers', 'enablePeerswap',
+    'lnServerUrl', 'swapServerUrl', 'boltzServerUrl', 'bitcoindConfigPath', 'channelBackupPath'
+];
+const NODE_AUTH_ALLOWLIST = ['swapMacaroonPath', 'boltzMacaroonPath'];
+const indexKey = (node) => +node.index;
+const isValidServerUrl = (url) => {
+    if (typeof url !== 'string' || url.trim() === '') {
+        return false;
+    }
+    try {
+        const parsed = new URL(url);
+        return (parsed.protocol === 'http:' || parsed.protocol === 'https:') && !!parsed.hostname;
+    }
+    catch {
+        return false;
+    }
+};
 // Set local block explorer URL after first API call
 // if the selected node block explorer has working REST API suite
 // otherwise set it to mempool.space
@@ -276,7 +299,19 @@ export const updateApplicationSettings = (req, res, next) => {
     const RTLConfFile = common.appConfig.rtlConfFilePath + sep + 'RTL-Config.json';
     try {
         const oldConfig = JSON.parse(fs.readFileSync(RTLConfFile, 'utf-8'));
-        const config = common.addSecureData(JSON.parse(JSON.stringify(req.body)));
+        // Allowlist the per-node payload before addSecureData runs, so an injected credential
+        // path, runeValue/options or unknown setting can never reach the runtime config or
+        // the file — the same strategy updateNodeSettings applies to its settings merge.
+        const requestBody = JSON.parse(JSON.stringify(req.body));
+        requestBody.nodes?.forEach((node) => {
+            if (node.authentication && typeof node.authentication === 'object') {
+                node.authentication = Object.fromEntries(Object.entries(node.authentication).filter(([key]) => NODE_AUTH_ALLOWLIST.includes(key)));
+            }
+            if (node.settings && typeof node.settings === 'object') {
+                node.settings = Object.fromEntries(Object.entries(node.settings).filter(([key]) => NODE_SETTINGS_ALLOWLIST.includes(key)));
+            }
+        });
+        const config = common.addSecureData(requestBody);
         const runtimeConfig = oldConfig;
         Object.keys(config).forEach((key) => {
             if (key !== 'nodes') {
@@ -285,10 +320,10 @@ export const updateApplicationSettings = (req, res, next) => {
         });
         if (config.nodes && config.nodes.length > 0) {
             const oldNodes = (common.appConfig.nodes && common.appConfig.nodes.length > 0) ? common.appConfig.nodes : (oldConfig.nodes || []);
-            const newNodesMap = new Map(config.nodes.map((node) => [node.index, node]));
+            const newNodesMap = new Map(config.nodes.map((node) => [indexKey(node), node]));
             const updatedAndExistingNodes = oldNodes.map((oldNode) => {
-                const newNode = newNodesMap.get(oldNode.index);
-                newNodesMap.delete(oldNode.index);
+                const newNode = newNodesMap.get(indexKey(oldNode));
+                newNodesMap.delete(indexKey(oldNode));
                 const node = newNode ? {
                     ...oldNode,
                     ...newNode,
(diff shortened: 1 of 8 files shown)


saubyk:
Thanks for tracing this properly and for asking before implementing — the three you picked out are exactly the load-bearing ones.

**1. `lnServerUrl` — yes, pin it, but you can drop the validation half.**

Pinning `lnServerUrl`/`swapServerUrl`/`boltzServerUrl` server-side for existing nodes is right. The settings UI never edits them: they appear in `src/app` only in the config model, the store state and test data, never as a form control. `app-settings.component.ts` just echoes the whole `appConfig` back from `getConfig`, so a client can never hold a legitimately new value for them.

The shape validation for new nodes becomes dead code under (2) below — no reachable caller — so `isValidServerUrl` can come out entirely.

One thing to fix while you're in there: `addSecureData` decides new-vs-existing from `this.appConfig.nodes`, but `updateApplicationSettings` decides the same question from `oldNodes`, which falls back to the file when `appConfig.nodes` is empty. Two sources for one decision. Please make it one.

**2. New nodes — stripping the credential paths is correct but doesn't finish the job.**

The four `authentication.*` fields aren't the only sinks. On the new-node path these stay caller-controlled and each reaches one:

- `settings.bitcoindConfigPath` → `getConfig('bitcoind')` reads the file and returns its contents (`RTLConf.ts:175`)
- `settings.channelBackupPath` → this *is* the containment root for `getFile` (`RTLConf.ts:84-95`), so a caller-chosen root defeats that check outright
- `authentication.swapMacaroonPath` + `settings.swapServerUrl` → `common.ts:149-151` reads `<path>/loop.macaroon` and sends it as a header to `swapServerUrl`. That's the original confused deputy rebuilt out of allowlisted parts. `boltzMacaroonPath`/`boltzServerUrl` are the same shape at `common.ts:167-169`
- `settings.logFile` → `fs.appendFile`, five sites in `logger.ts`

Note this is also true for *existing* nodes in the swap/boltz case, and the new test encodes it — `assert.equal(existingNode.authentication.swapMacaroonPath, '/loop/swap')` asserts that an attacker-supplied path survives.

Nothing fires in the running process, since these read from `common.nodes` and this endpoint only reassigns `common.appConfig`. But the node is persisted to `RTL-Config.json` and `config.ts:245-266` loads exactly those fields back into `common.nodes` at the next restart. Deferred, not prevented — and it persists.

So: **don't create nodes through this endpoint at all.** Drop unknown-index nodes and log a WARN rather than provisioning them. There's no add-node UI — your own justification for stripping credentials — and this collapses the entire class instead of maintaining a strip list that drifts as fields are added.

Separately, `logFile` should come out of the settings allowlist regardless: `config.ts:286` overwrites it unconditionally at boot from `rtlConfFilePath`, so accepting it buys nothing.

**3. `secret2FA` — yes, intended, and the test change is fine.**

I checked the rotate path: `onVerifyToken()` disables first (empty seed + `enable2FA` false, which your rule permits), then a second request sets the fresh seed while the server seed is empty, which your `else if` honors. There's no single-request reseed, so nothing legitimate breaks. The old assertion encoded the behavior we're removing, and you've added a test for the enable case — that's the right way to change it.

Two smaller things: `NODE_ALLOWLIST` drops top-level `logLevel`, `logFile` and `lnVersion`, which `SelectedNode` declares at node level — harmless today because the merge preserves the old values, but please confirm the UI doesn't send a top-level `logLevel` change. And the release note landed in `Release-notes-0.15.11.md`; 0.15.11 has shipped, so it needs to move to `0.15.12`.

saubyk:
On the release note — `release-notes/Release-notes-0.15.12.md` already exists on the release branch, so it just needs a new `## Bug Fixes` section rather than a new file. The entry itself reads well and is in the right format; keep the `#TBD` placeholder and we'll fill in the number after merge.

Heads-up that the last two clauses ("strips credential paths from any node the server does not already know, and rejects a malformed `lnServerUrl` on such a node") will need rewording once (2) above is addressed — if unknown-index nodes are dropped rather than provisioned, neither behavior exists any more.

7-falseparfait:
Thanks for the thorough review @saubyk. 
 Pushed the fixes:

1 . Unknown-index nodes are dropped with a WARN instead of provisioned; nothing caller-supplied gets persisted for pickup at restart anymore.
2  `isValidServerUrl` removed as dead code, and known-vs-unknown comes from a single source (`common.appConfig.nodes`) in both `addSecureData` and the merge.
3 Swap/boltz macaroon paths pinned server-side too; legit edits still go through /node, which is untouched.
4 logFile out of the allowlist; also confirmed the UI never sends a top-level node logLevel.
5 Release note moved to Release-notes-0.15.12.md, reworded per your heads-up.

Backend rebuilt, all backend tests passes

7-falseparfait:
F1 addressed in `6cb67ab8`. The `blockExplorerUrl` value is now validated at the controller level before it reaches request(options): it must parse as a valid URL with an http: or https: scheme; anything else is silently dropped. This is format validation only; it rejects malformed strings and non-HTTP schemes, but does not prevent an authenticated caller from pointing at an internal host. That is a deliberate trade-off: a user-chosen block explorer (including self-hosted mempool instances) is an intended RTL feature, and pinning the URL to the server-held value would break it. The PR description has been updated to reflect that the validation is format-level, not SSRF-prevention as this wwill break self hosted mempool instances and the likes

saubyk:
Pushed some more fixes and merged. Thanks for the effort @7-falseparfait

7-falseparfait:
pheww thanks @saubyk  appreciate you stepping in with the last round of fixes appreciate it

Issue #1689, opened 2026-08-29
Docker fixture: Eclair pinned at 0.13.1, current release is 0.14.2 (needs bitcoind 31)


saubyk:
The `docker/` regtest fixture pins `polarlightning/eclair:0.13.1` (`docker/docker-compose.yml`). Eclair's current release is v0.14.2 (2026-08-26), so Eclair changes are being tested against a version one minor behind.

### Why it isn't a one-line bump

- **No multi-arch 0.14.x image is published.** Polar's newest tag is still `0.13.1` (Nov 2025). `acinq/eclair:latest` was rebuilt 2026-08-28 but is amd64-only, which is why the fixture uses Polar's build in the first place (see the comment above the `eclair` service).
- **Eclair 0.14.1+ requires Bitcoin Core 31.x.** The fixture runs `polarlightning/bitcoind:30.0`, shared with the three LND nodes and Core Lightning, so the bitcoind bump has to be validated against all of them.
- Building Eclair from source inside the fixture would add a JDK/Maven build stage — extra supply-chain surface for a dev-only fixture, so not the preferred route.

### Compatibility notes from the 0.14.x release notes

- 0.14.0 removed the `channelstats` and `networkfees` APIs (replaced by `relaystats`) and reworked the AuditDb (pre-0.14 audit rows are moved to `_before_v14` tables and no longer visible via `/audit`).
- RTL still has `getChannelStats` in `server/controllers/eclair/channels.ts` calling `/channelstats`, but it has no route and no frontend caller — dead code, not a live break. Worth removing when that file is next touched.
- Everything else RTL calls (`listinvoices`, `listpendinginvoices`, `getreceivedinfo`, `audit`, `channels`, `peers`, `payinvoice`, …) is unchanged through 0.14.2 ("API changes: nothing noteworthy").

### Options

1. Wait for a Polar multi-arch `0.14.x` tag, then bump `eclair` + `bitcoind` together.
2. Switch the `eclair` service to `acinq/eclair` with `platform: linux/amd64` and accept emulation on arm64 hosts.
3. Add a source build stage (least preferred).

Until then, Eclair work (e.g. #1067) is tested against 0.13.1 and should say so in the PR.

PR #1690, opened 2026-08-29
Page Eclair invoices server-side instead of fetching the whole table

### LINKED ISSUE #1067: Is there a way to limit the API request rate?
### Discussed in https://github.com/Ride-The-Lightning/RTL/discussions/1066

<div type='discussions-op-text'>

<sup>Originally posted by **DerEwige** August 16, 2022</sup>
My RTL freezes (overloads) my eclair node when it tries to load the invoices.
(When I go to lightning -> transactions)

The node CPU goes up to 100% (4 cores) and I get time out errors in the logs for http://localhost:8080/listinvoices

I asume that RTL has problem handling over a million of invoices.
Is there an option in the config to tell RTL to load the data slower?
</div>

saubyk:
Fixes #1067.

## Problem

Opening the Eclair dashboard or Transactions page asked the node for every invoice it had ever issued (`count=1000000`) and then called `/getreceivedinfo` once per invoice, all concurrently. On a node with hundreds of thousands of invoices that pinned Eclair at 100% CPU until the request timed out and RTL retried.

Eclair's `listinvoices` has had `count`/`skip` since 2022, but it returns oldest-first and never reports a total, and the request for one (ACINQ/eclair#2855) was closed — so the paging plumbed in #1393 stayed switched off.

## Change

- **Backend** (`server/controllers/eclair/invoices.ts`): derives the total itself with single-row probes (gallop, then bisect: O(log n) calls that each cost Eclair one `LIMIT 1 OFFSET n` query) and serves one newest-first page as `{ invoices, totalInvoices }`. Page size defaults to 10 and is capped at 100 whatever the client sends, so the per-invoice status fan-out is bounded by the page. The separate unbounded `listpendinginvoices` call is dropped — `/getreceivedinfo` already reports `pending`. Malformed `count`/`skip` are refused with a 400 instead of forwarded.
- **Frontend**: the invoices table pages server-side with an exact total; sort and the filter box apply within the current page. The transactions report, which summed received amounts from the full invoice list, now uses the audit's received payments (already loaded by the payments fetch).
- Release-note entry under `## Bug Fixes` (`#TBD` to be filled in).

## Verification

- `npm run lint` ✅, `npm run test` ✅ (backend 73/73 incl. 7 new in `test/backend/eclair-invoices.test.mjs`, written first and failing; Karma 230/230 incl. a new reports spec).
- Docker fixture, Eclair 0.13.1, 57 invoices with 3 paid from bob: API drive — newest-first pages, ragged last page, `count=1000000` capped, 400 envelope on `count=abc`, one `/getreceivedinfo` per invoice on the page. Headless UI: paginator reads "11 – 20 of 57" / "51 – 57 of 57"; reports show "Received 168 Sats/3 Invoices" (= 55+56+57).
- The fixture's Eclair is one minor behind the current release (0.14.2); tracked in #1689. The three endpoints touched here are unchanged through 0.14.2.

## Notes

- The Payments page still fetches the audit with `count=1000000`; same shape, deliberately out of scope here — follow-up issue material.
- `getChannelStats` in `server/controllers/eclair/channels.ts` calls `/channelstats`, removed in Eclair 0.14.0, but has no route or caller (dead code) — noted in #1689.


saubyk:
**Review conclusion from Kimi K3 (AI reviewer) — approve.**

Reviewed the full source diff, then independently re-ran every check locally on the PR head (`c9026cac`):

**Artifact verification (the one gate not reviewable by reading the diff)**
- `npm run buildbackend` → zero diff against committed `backend/`
- `npm run buildfrontend` → zero diff against committed `frontend/`; every chunk hash matches (`main.41edb45745ffe0ea.js`, `runtime.9b45958eeae4c77e.js`, `17.47e2ffe55b539e16.js`). The committed bundles provably came from these sources.

**Checks**
- `npm run lint` — clean
- Backend: **73/73 pass**, including the 7 new `test/backend/eclair-invoices.test.mjs` specs
- Karma: **231/231 SUCCESS** (230 on base + the new transactions-report spec)

**Source review highlights**
- Gallop/bisect total derivation is correct on paper (termination, off-by-one, empty-node case) and each probe is a single-row `LIMIT 1 OFFSET n` — the #1067 invariant (no unbounded `/listinvoices`, no `/listpendinginvoices` from `listInvoices`, status fan-out bounded by page) is enforced and pinned by behavior-level tests that fail on the old code.
- Newest-first window mapping (`pageSkip = total - skip - pageCount` + reverse), ragged last page, and skip-beyond-total are all handled and tested.
- Frontend wiring is consistent: exact paginator total, `onPageChange` correctly uses `event.pageSize`, dropping `this.invoices.paginator` is required for server-side paging, and `createNewInvoice` bumping `totalInvoices` keeps the paginator honest.
- Units check out: `fees.ts` already converts audit `part.amount` msat→sats, and `firstPartTimestamp / 1000` matches the report filter's seconds.

**Non-blocking observations (follow-up material, not for this PR)**
1. The filter box and CSV download now operate within the current page only (previously the fully-loaded list).
2. Report "invoices" may now count non-invoice receipts (e.g. keysend), since the audit's received payments include them — arguably more accurate, but a behavior change.
3. `circularRebalance` in `server/controllers/eclair/channels.ts` still calls `/listpendinginvoices` unbounded — same shape as #1067 for a node with a huge pending set; belongs with the payments `count=1000000` follow-up.

**Environment note:** the stock `npm run test` initially failed on this machine — system Chrome 151's new headless binary breaks Karma's socket transport (connects, freezes at "Executed 0 of N", ping-timeout disconnect). This reproduces identically on the base branch, so it is not a PR regression. Pointing `CHROME_BIN` at Playwright's `chrome-headless-shell` (also 151) runs the full suite clean. This will hit every local `npm run test` on Chrome ≥151 — worth pinning headless-shell or documenting `CHROME_BIN` separately.

Issue #1656, opened 2026-08-03
Login lockout counter: broken expiry, unbounded growth, and X-Forwarded-For keying


saubyk:
Follow-up from the review on #1654 (finding F5). These are **pre-existing** issues in the login lockout logic in `server/controllers/shared/authenticate.ts`, deliberately kept out of that PR's scope.

## 1. Lockout never expires (permanent lockout)

`getFailedInfo` tests expiry against a freshly-initialized local `failed` whose `lastTried === currentTime`, so the comparison `currentTime > (failed.lastTried + LOCKING_PERIOD)` is always false and an existing entry is never treated as expired. The only other cleanup is the `loginInterval` sweeper, which calls `clearInterval(loginInterval)` **inside its own loop** — the first pass that deletes anything kills the sweeper for the process lifetime.

Net effect: after 5 failed attempts an IP stays locked out until the process restarts, even with the correct password (and after the nominal 30-minute period the error message degrades to "locked after 0 more attempts"). An attacker can therefore deny the legitimate operator access indefinitely.

## 2. Unbounded state growth

`failedLoginAttempts` is a plain object keyed by request IP with no size bound. With (1) breaking cleanup, entries accumulate for the process lifetime — a memory-growth vector given attacker-controlled keys (see 3). Plain-object keying also has `__proto__`/`constructor` hazards; a `Map` or `Object.create(null)` is the right structure.

## 3. Key is client-controllable

The counter keys on `common.getRequestIP(req)`, which trusts `X-Forwarded-For` first (`app.set('trust proxy', true)`). Rotating the header yields a fresh counter per request, so the 5-attempt lockout can be bypassed outright by a brute-forcing client.

Fixing this direction needs a maintainer decision on trusted-proxy configuration, since RTL legitimately runs behind reverse proxies — e.g. pin `trust proxy` to a hop count, key on the socket peer absent trusted-proxy config, and/or add a global (not per-key) attempt budget.

## Suggested scope for the follow-up

- (1) and (2) are straightforward: fix the expiry comparison to use the stored entry, remove the stray `clearInterval`, and switch to a `Map`.
- (3) needs the proxy-handling decision above.
- Regression tests in `test/backend/authenticate.test.mjs` (the harness wired up in #1654).

saubyk:
[...]

PR #1691, opened 2026-08-30
Fix login lockout expiry and bound the failed-attempt table

saubyk:
Part of #1656 (findings 1 and 2). Finding 3 — the counter keys on `X-Forwarded-For` — needs a trusted-proxy decision and is left for a follow-up.

## Root cause

- `getFailedInfo` compared the clock against a freshly created local entry (`lastTried === currentTime`) rather than the stored one, so the 30-minute expiry never fired.
- The sweeper called `clearInterval` on itself inside its own loop, so the first deletion stopped it for the process lifetime.
- Net effect: an address that hit five failed attempts stayed locked until restart, even with the correct password.
- Counters lived in a plain object keyed by client-supplied address: no size bound, and `__proto__`/`constructor` key hazards.

## Change

- Expiry is tested against the stored entry; the sweeper keeps running.
- Counters live in a `Map` capped at 1000 addresses (oldest entry evicted on insert).
- `LOCKING_PERIOD`, `ALLOWED_LOGIN_ATTEMPTS`, `MAX_TRACKED_ADDRESSES` exported for the tests.

## Verification

- Four new tests in `test/backend/authenticate.test.mjs` (expiry via `getFailedInfo`, end-to-end unlock through `authenticateUser`, bounded table, prototype-name keys) — all failed before the fix, pass after.
- `npm run testbackend`: 77/77 pass. `npm run lint`: clean.
- No `src/` change, so the frontend suite was not run; `backend/` regenerated.

### DIFF
--- a/backend/controllers/shared/authenticate.js
+++ b/backend/controllers/shared/authenticate.js
@@ -38,18 +38,16 @@ export const getFailedInfo = (reqIP, currentTime) => {
     const existing = failedLoginAttempts.get(reqIP);
     return (existing && !hasExpired(existing, currentTime)) ? existing : { count: 0, lastTried: currentTime };
 };
-export const recordFailedAttempt = (reqIP, failed, currentTime) => {
-    // Guard the stored entry, not the argument: a live lockout is immutable whatever the
-    // caller holds. The argument only seeds the table when nothing live is stored for reqIP.
-    const stored = failedLoginAttempts.get(reqIP);
-    if (stored && !hasExpired(stored, currentTime)) {
-        failed = stored;
-    }
+// Records one failure for reqIP and returns the entry now stored for it. Derives the
+// entry itself rather than trusting a caller-held object, so a live lockout is immutable
+// and an expired one is dropped whatever the caller holds.
+export const recordFailedAttempt = (reqIP, currentTime) => {
+    const failed = getFailedInfo(reqIP, currentTime);
     // A locked address records nothing more: the 401 it gets must depend on the lockout
     // alone, never on the credential offered (a moving deadline would tell a wrong guess
     // from a right one), and the window is fixed rather than extendable by more failures.
     if (isLocked(failed, currentTime)) {
-        return;
+        return failed;
     }
     failed.count = failed.count + 1;
     failed.lastTried = currentTime;
@@ -70,6 +68,7 @@ export const recordFailedAttempt = (reqIP, failed, currentTime) => {
         failedLoginAttempts.delete(evict);
     }
     failedLoginAttempts.set(reqIP, failed);
+    return failed;
 };
 const handleMultipleFailedAttemptsError = (failed, currentTime, errMsg) => {
     if (isLocked(failed, currentTime)) {
@@ -150,8 +149,7 @@ export const authenticateUser = (req, res, next) => {
             if (common.appConfig.enable2FA && common.appConfig.secret2FA && common.appConfig.secret2FA !== '' && !hasValidAuthToken(req)) {
                 if (typeof twoFAToken !== 'string' || twoFAToken === '' || !verifyToken(twoFAToken)) {
                     logger.log({ selectedNode: req.session.selectedNode, level: 'ERROR', fileName: 'Authenticate', msg: 'Invalid Token! Failed IP ' + reqIP, error: { error: 'Invalid token.' } });
-                    recordFailedAttempt(reqIP, failed, currentTime);
-                    return res.status(401).json(handleMultipleFailedAttemptsError(failed, currentTime, 'Invalid 2FA Token!'));
+                    return res.status(401).json(handleMultipleFailedAttemptsError(recordFailedAttempt(reqIP, currentTime), currentTime, 'Invalid 2FA Token!'));
                 }
             }
             if (!req.session.selectedNode) {
@@ -163,11 +161,11 @@ export const authenticateUser = (req, res, next) => {
             res.status(200).json({ token: token });
         }
         else {
-            logger.log({ selectedNode: req.session.selectedNode, level: 'ERROR', fileName: 'Authenticate', msg: 'Invalid Password! Failed IP ' + reqIP, error: { error: 'Invalid password.' } });
-            if (common.appConfig.rtlPass !== password) {
-                recordFailedAttempt(reqIP, failed, currentTime);
-            }
-            return res.status(401).json(handleMultipleFailedAttemptsError(failed, currentTime, 'Invalid Password!'));
+            // Reached for a wrong password, or for a correct one refused by a live lockout.
+            const wrongPassword = common.appConfig.rtlPass !== password;
+            logger.log({ selectedNode: req.session.selectedNode, level: 'ERROR', fileName: 'Authenticate', msg: (wrongPassword ? 'Invalid Password! Failed IP ' : 'Locked Out! Failed IP ') + reqIP, error: { error: wrongPassword ? 'Invalid password.' : 'Address locked out.' } });
+            const recorded = wrongPassword ? recordFailedAttempt(reqIP, currentTime) : failed;
+            return res.status(401).json(handleMultipleFailedAttemptsError(recorded, currentTime, 'Invalid Password!'));
         }
     }
 };
--- a/backend/controllers/shared/authenticate.js
+++ b/backend/controllers/shared/authenticate.js
@@ -7,15 +7,20 @@ import { Common } from '../../utils/common.js';
 const logger = Logger;
 const common = Common;
 const ONE_MINUTE = 60000;
-const LOCKING_PERIOD = 30 * ONE_MINUTE; // HALF AN HOUR
-const ALLOWED_LOGIN_ATTEMPTS = 5;
-const failedLoginAttempts = {};
+export const LOCKING_PERIOD = 30 * ONE_MINUTE; // HALF AN HOUR
+export const ALLOWED_LOGIN_ATTEMPTS = 5;
+export const MAX_TRACKED_ADDRESSES = 1000;
+// Keyed by request IP. A Map (not a plain object) so client-supplied keys such as
+// "__proto__" cannot collide with Object.prototype, and bounded so an attacker
+// rotating addresses cannot grow it without limit (issue #1656).
+const failedLoginAttempts = new Map();
 const databaseService = Database;
+const hasExpired = (failed, currentTime) => currentTime > (failed.lastTried + LOCKING_PERIOD);
 const loginInterval = setInterval(() => {
-    for (const ip in failedLoginAttempts) {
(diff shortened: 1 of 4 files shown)


saubyk:
*Reviewed by Kimi K3 (automated code review — K3).*

**Approve.** Root-cause analysis is correct, the fix is minimal and matches the file's style, the tests pin the exact regressions, and every claim in the PR body that can be checked locally checks out.

### Verified (not just read)

- `npm run testbackend` — **77/77 pass**, including the 4 new tests.
- `npm run lint` — clean.
- `npm run buildbackend` → `git status` clean — the committed `backend/` is a faithful recompile of `server/`, so there's nothing smuggled in the build output.
- **Reproduced the bugs against the pre-PR compiled code** (swapped in `070d9620^`'s `authenticate.js`, probed, restored):
  - An entry at 5 failures still read `count: 5` 31 minutes later — expiry never fired, lockout lasted until restart. ✔ bug confirmed
  - `getFailedInfo('__proto__')` returned `Object.prototype` itself (`.count` undefined) — the plain-object key hazard was real, not theoretical. ✔
- Release note is under `## Bug Fixes`, links the PR and #1656, states root cause, and the `#TBD` → `(#1691)` follow-up commit follows the repo workflow.
- `failedLoginAttempts`/`getFailedInfo` have no consumers outside this module and the test, and there is no parallel copy of this logic under `lnd/`/`cln/`/`eclair/` — `shared/` is the only place it lives.

### Code-level check

- `hasExpired` (strict `>`) and `handleMultipleFailedAttemptsError` (`<=`) agree on the boundary — locked at exactly `lastTried + LOCKING_PERIOD`, unlocked 1 ms later, and the first new test pins that.
- Deleting from the `Map` inside the sweeper's own iteration is spec-safe.
- Cap logic is sound: expired-but-present keys are overwritten in place (no size change); absent keys evict-then-insert only at `size >= 1000`, so the map can never exceed the cap.
- The bounded-table test is robust to leftover module state from earlier tests — leftovers are always older than `bound-first`, and exactly that many evictions precede it.
- Eviction is fail-open (an attacker rotating >1000 addresses can clear *other* addresses' lockouts early), but that grants nothing without the password, and rotation-based bypass is finding 3, explicitly deferred. Right call.

### Nits (non-blocking, no change requested)

- `Map.set` on an expired existing key keeps its original insertion position, so eviction order is first-insert rather than last-touch. Irrelevant at 1000 entries / 30-minute expiry.
- The bounded test's message says "the oldest entry" when, with leftovers, it's strictly "the oldest of the `bound-*` entries." Cosmetic.

### One unverified item

The repo guidance is to run `docker/scripts/verify-sso.sh` after touching authentication. I couldn't — the docker daemon wasn't running. Risk looks nil (the change is confined to the password-branch counter bookkeeping; the SSO branch never reaches `getFailedInfo`), but worth a pass of the SSO harness before merge if you want the letter of the gate satisfied. Everything else is green.

saubyk:
Thanks — closing out the two open items:

- **SSO harness**: brought the fixture up with `--profile sso`, copied this branch's `backend/` into `rtldev_rtl_sso`, ran `docker/scripts/verify-sso.sh` → **11 passed, 0 failed**. Also drove the standalone RTL's password login over the API: four "locked after N more attempts" warnings, then the lockout message on the fifth and sixth — as expected on the new code.
- **Eviction-order nit**: taken as a real (if minor) fix rather than left. `getFailedInfo` now deletes before it sets, so a counter reset after expiry moves to the back of the Map and the cap evicts the least recently reset entry. Pinned by a new test that fails without the delete line (checked by removing it from the compiled output). The bounded-table assertion message is reworded as suggested.

`npm run testbackend` 78/78, `npm run lint` clean.

saubyk:
All three findings taken, in the latest commit.

**F2 (major, eviction dropped live lockouts)** — agreed, the policy was backwards. Two changes:
- `getFailedInfo` is now read-only; only `recordFailedAttempt` inserts. First-contact visitors and successful logins no longer consume a slot (and a successful login can no longer evict someone else's lockout).
- A full table evicts the least recently failed address that is **not** currently locked, falling back to the oldest lockout only when every tracked entry is locked. Pinned by two tests: a lockout at the head of the table survives 1010 other addresses failing once each, and the all-locked fallback evicts the oldest.

**F1 (sweeper untested)** — the callback is extracted as `sweepExpiredAttempts(now)`. The test inserts two expired entries and one live one, sweeps once, and asserts the table size via a new `trackedAddresses()` accessor — reading through `getFailedInfo` would have passed either way because of the lazy expiry, which is exactly the gap you pointed at.

**F3 (saturated shared state)** — `clearFailedAttempts()` is exported and every test that fills the table calls it first (the last one also clears after itself), so no test depends on what ran before it.

Each new test was checked against the old behaviour by patching the compiled `backend/` (head-only eviction; sweeper returning after its first delete) — the corresponding test fails, the rest pass. Release note updated to describe the eviction policy.

`npm run testbackend` 81/81, `npm run lint` clean, `verify-sso.sh` 11/11 against this build, and the live fixture still locks on the fifth wrong password.

saubyk:
All five taken in the latest commit.

**F4 (blocker) / F5** — agreed on both, and thanks for catching that my comment had blessed the asymmetry. The guard lives inside `recordFailedAttempt` rather than at the call site: on a locked entry it is a no-op, so the policy has one home and the 2FA branch inherits it too. Two new tests, both confirmed failing before the change: a wrong guess at `LOCKING_PERIOD - 1` neither moves `lastTried` nor grows `count`, and the original deadline still lifts the lock; and, through `authenticateUser`, a right and a wrong password against a live lockout return byte-identical 401 bodies with `lastTried` untouched (the stored deadline is shifted by 10 minutes first so a leak would be visible). Verified live on the fixture as well: wrong → right → wrong while locked all render the same `Try again after` timestamp, and a fresh address still gets a token.

**F6** — you're right, and the correction stands against my earlier comment too: `clearInterval` cancels later fires without breaking the loop, so my "regression proof" patched in a stricter behaviour than the one that shipped. Comment reworded per your suggestion; the test now claims only what it drives (one pass clears every expired entry). The release note says "sweep pass" rather than "sweeper".

**F7** — comment now states the aliasing contract: the live table entry when one exists (mutating it mutates the table), a detached object otherwise, tracked only once handed to `recordFailedAttempt`.

**F8** — `handleMultipleFailedAttemptsError` calls `isLocked` instead of re-deriving it.

`npm run testbackend` 83/83, `npm run lint` clean, `verify-sso.sh` 11/11 against this build.

saubyk:
F8, F10 and F11 taken; pushing back on F9; F12 noted as out of scope.

**F9 — disagree that this is exploitable, and the proposed fix is worse than the hole.**
The scan evicts the first *unlocked* entry from the head, and it only runs when the table is full. So to flush a count-4 counter the attacker first needs the other 999 slots occupied, and then needs their own entry to be the oldest unlocked one — i.e. every filler entry must have been (re)inserted after it. Fillers are not free: each filler failure increments that address's own counter, and after four it locks, at which point the F4 guard means it never moves again and the scan skips it. So each filler address contributes at most four re-insertions, and one reset of the attacker's counter costs on the order of 250 fresh addresses. But 250 fresh addresses used *directly* give ~1000 guesses. The trick buys four guesses for what direct use turns into a thousand: it is roughly 250× worse for the attacker, not better. (The all-locked fallback is the same shape at a higher price — 1000 locked addresses — and yields one eviction per new address, again strictly worse than just using the address.) If the key is attacker-chosen, the deferred `X-Forwarded-For` finding gives five guesses per request with no table games at all; if it is not, an attacker with ≥1000 real addresses already has 4–5 guesses each per window, and this policy does not add to that.

The suggested alternative — evict only expired entries and fail closed (reject/429) when none exist — converts that non-gain into a clean denial of service: 1000 addresses, or 1000 header values, fill the table with unexpired entries and every login is refused for 30 minutes, the operator's included. That is the "attacker denies the legitimate operator access" outcome #1656 opened with, reintroduced by design. A bounded table that fails open on the entries that matter least is the right trade here, and it stays that way until the keying decision closes finding 3. Comment at the eviction site left as is; happy to add the argument above if you'd like it in the code.

**F10** — taken. `recordFailedAttempt` now re-reads the table for `reqIP` and applies the locked check to the stored entry; the argument only seeds the table when nothing live is stored. Test: a stale detached object for a locked address does not replace the lockout (failed before, passes after).

**F11** — taken. New test: four failures, a correct login, then `trackedAddresses()` is 0 and the count reads 0. Checked that it fails with the success-path `delete` removed from the compiled output.

**F8 (remaining site)** — taken. The success gate uses `!isLocked(failed, currentTime)`.

**F12** — agreed it is real (`timingSafeEqual` throws on a length mismatch, and `app.ts` does have a catch-all error handler at `:69`, so it's a 500 not a crash), but it is the SSO branch, untouched by this PR, and belongs in its own issue.

`npm run testbackend` 85/85, `npm run lint` clean, `verify-sso.sh` 11/11 against this build.

saubyk:
F10 and F14 taken; F13 deliberately left for a separate decision.

**F10** — taken in the structural form suggested: `recordFailedAttempt(reqIP, currentTime)` no longer accepts an entry. It looks the entry up itself, so an expired counter cannot be revived by a caller-held object and there is no argument to pass `undefined` for, and it returns the entry it stored, which is what both 401 sites now render. Test rewritten to pin both halves: on a locked address the returned entry is the live one, unchanged; after expiry the returned entry has `count: 1` and is identical to what `getFailedInfo` reads back.

**F14** — taken; a correct password refused by a live lockout now logs `Locked Out! Failed IP …` rather than `Invalid Password!`. Confirmed on the fixture: the fifth wrong password logs invalid password, the correct one that follows logs locked out.

**F13** — not changed here, on purpose. It's real and pre-existing: below the threshold the two 401 bodies differ (`Invalid Password!` vs `Invalid 2FA Token!`), so a caller holding the password can confirm it without a second factor, bounded by the five-attempt counter. Collapsing them to one generic body is standard practice, but it changes what operators see on the login screen (they lose "which field was wrong") and touches the existing 2FA assertions, so it's a maintainer call rather than something to fold into a lockout-bookkeeping fix. I'd suggest tracking it alongside the `X-Forwarded-For` keying under #1656, since the counter that bounds it is the one whose keying is deferred.

`npm run testbackend` 85/85, `npm run lint` clean, `verify-sso.sh` 11/11 against this build.

saubyk:
**F13** — taken after all, on the maintainer's call. When a second factor is required for the request, both failure replies now read `Invalid Password or 2FA Token!`; the distinction is kept in the server log only. Without 2FA (or for an already-authenticated session re-authorizing by password, which needs no token) the message is `Invalid Password!` as before. No existing assertion had to change — the `/2FA/` and `/Invalid Password/` matches still hold — and a new test asserts that a wrong password, a right password with a wrong token, and a right password with no token all return the same body (failed before, passes after). Release note updated.

`npm run testbackend` 86/86, `npm run lint` clean, `verify-sso.sh` 11/11 against this build.

PR #1695, opened 2026-09-03
Report a failed Eclair channel open instead of success and Unknown Error

stupleb:
Eclair answers /open with HTTP 200 whatever the outcome; only a body starting with 'created channel' means a channel was opened. Treat everything else as the error it is, carrying Eclair's reason. Guard the channel mapper against a channel that has no commitments yet, which Eclair lists while an open is in flight and for a minute after one fails, so the list no longer collapses into Unknown Error.


saubyk (approved):
LGTM

PR #1696, opened 2026-09-07
Configure Loop and Boltz connection settings outside the UI

saubyk:
Follow-up to #1683, which pinned the LN server URL and credential paths so they can only come from `RTL-Config.json` or the environment. The Loop and Boltz server URLs and macaroon directories are the same kind of setting: RTL reads a macaroon from the configured directory and sends it to the configured URL. This PR applies the same rule to them.

**Changes**

- The Services tab under Node Config is removed, along with its Loop, Boltz, Peerswap (already disabled) and "No Service" panels, their route and module declarations.
- `updateNodeSettings` no longer accepts `swapServerUrl` or `boltzServerUrl` in `settings`, and ignores `req.body.authentication` entirely. `filterNodeSettings` loses its service-URL mode; `applyWritableServiceMacaroonPaths` is gone.
- The Loop and Boltz feature pages are unchanged and still show whenever the corresponding server URL is set, via config or the `SWAP_*` / `BOLTZ_*` env vars, which `config.ts` already honours.
- Release note entry under Enhancements.

**Tests**

- Backend: the two `/node` tests now assert every server URL is pinned and every `authentication` field is ignored (including an attempt to clear a path with an empty string). 108/108 pass.
- Frontend: 230/230 (the deleted Services spec accounts for the drop from 231). `npm run lint` green.
- `backend/` and `frontend/` regenerated from these sources.


saubyk (commented):
**Review — Kimi K3 (AI reviewer, posted via maintainer account)**

Two nits from a pass over head `db0a55264`, neither blocking.

**1. `addSecureData` still pins `swapMacaroonPath`/`boltzMacaroonPath`** (`backend/utils/common.js`, mirrored in `server/utils/common.ts`). This is **correct** — the application-settings endpoint can still receive those fields in a request body, so the pin is the defense-in-depth layer that keeps them server-held. Just noting it's intentional, not an oversight, and the updated comment there ("no endpoint edits them") is now accurate.

**2. Reported test counts are unverified** (108 backend / 230 frontend / lint green). Nothing runs CI on an open PR, so `npm run lint && npm run test` should be run on this head before merge as the only gate.

And one pre-merge check: the `db0a55264` "Fill in PR number" commit touched only the release note, so if `buildbackend`/`buildfrontend` were last run on `775c5d925` the committed artifacts should still be current — worth confirming the builds postdate the final source edit.

Otherwise LGTM

saubyk:
Both points addressed.

**1. `addSecureData` pin** — intentional. The application-settings handler strips `authentication` wholesale before `addSecureData` runs, and the pin stays as the defense-in-depth layer behind it; its comment ("no endpoint edits them") is accurate after this PR.

**2. Gate on the final head `109ea277`** — run just now, then both artifacts rebuilt from source and diffed against the committed copies:

| Check | Result |
|---|---|
| `npm run lint` | green |
| `npm run buildbackend` | clean |
| `npm run testbackend` | 108 pass, 0 fail |
| Frontend Karma (`ng test`) | 230 of 230 |
| `npm run buildfrontend` | clean |
| `git status backend/ frontend/` after both rebuilds | no drift |

The artifacts were originally built on `775c5d92`; the later commits touched only the release note and signatures, and the rebuild confirms the committed `backend/` and `frontend/` match the head sources.

Issue #1698, opened 2026-09-07
Remaining query-parameter concatenations send "undefined" upstream (follow-up to #1678)


saubyk:
Follow-up to #1678, which fixed the two list controllers (`listInvoices`, `getPayments`) in
#1687 but deliberately left the single-parameter sites of the same shape untouched. This issue
tracks the remainder.

### The defect class

A handler builds the upstream URL by concatenating `req.query.*` (or `req.params.*`) straight
into it, with no default, no validation and no encoding. Two consequences:

- An omitted parameter is interpolated as the literal string `"undefined"` and sent upstream,
  where LND's REST gateway cannot parse it as a `uint64`/`bool` and answers 400 — instead of
  the parameter simply being absent and LND applying its own default.
- A value containing `&` or `=` passes through into the upstream query string as extra
  parameters. The caller is an authenticated operator, so this isn't a privilege boundary, but
  it's the kind of thing validation at the edge should not permit.

### Remaining sites

**LND**

| File | Line | Parameter |
|---|---|---|
| `server/controllers/lnd/invoices.ts` | 34 | `?payment_addr=` (`invoiceLookup`) |
| `server/controllers/lnd/invoices.ts` | 36 | `?payment_hash=` (`invoiceLookup`) |
| `server/controllers/lnd/newAddress.ts` | 12 | `?type=` |
| `server/controllers/lnd/channels.ts` | 218 | `?force=` |
| `server/controllers/lnd/channels.ts` | 219 | `&target_conf=` |
| `server/controllers/lnd/channels.ts` | 220 | `&sat_per_vbyte=` |
| `server/controllers/lnd/wallet.ts` | 101 | `?max_confs=` |

**Shared (Loop)** — `server/controllers/shared/loop.ts`, lines 49, 68, 69, 124, 143 and 144:

```ts
options.uri = '/v1/loop/out/quote/' + req.params.amount + '?conf_target=' +
  (req.query.targetConf ? req.query.targetConf : '2') +
  '&swap_publication_deadline=' + req.query.swapPublicationDeadline;
```

`conf_target` has a default, so it is safe; `swap_publication_deadline` has none and hits the
`"undefined"` case exactly as #1678 described. `req.params.amount` is also interpolated into
the path unvalidated.

**Not affected:** the CLN and Eclair controllers — neither tree concatenates `req.query` or
`req.params` into a URL.

### Suggested fix

Same treatment #1687 applied to the list controllers: build the query through the `qs` option
[...]

PR #1709, opened 2026-09-20
Fix undefined query params in remaining LND and Loop controllers (#1698)

SIDHARTH20K4:
Follow-up to #1687, closes #1698.

Same fix applied to the remaining LND and Loop handlers that built the
upstream URL by concatenating query params:

- invoiceLookup (?payment_hash= / ?payment_addr=)
- getNewAddress (?type=)
- closeChannel (?force=, &target_conf=, &sat_per_vbyte=)
- getUTXOs (?max_confs=)
- loopOutQuote, loopInQuote, loopOutTermsAndQuotes, loopInTermsAndQuotes
  (&swap_publication_deadline=, and req.params.amount in the path)

Each handler now builds the query through options.qs, validates at the
boundary (non negative safe integers for counts and fee targets, boolean
spellings for force, the known enum for type), and omits a parameter
that is absent so the node or Loop server applies its own default.

Backend tests added for each handler. They assert no undefined appears
in the outgoing URL for an empty req.query, and that malformed input
returns 400.


server/controllers/lnd/newAddress.ts
```diff
@@ -9,8 +9,16 @@ export const getNewAddress = (req, res, next) => {
   logger.log({ selectedNode: req.session.selectedNode, level: 'INFO', fileName: 'NewAddress', msg: 'Getting New Address..' });
   options = common.getOptions(req);
   if (options.error) { return res.status(options.statusCode).json({ message: options.message, error: options.error }); }
-  options.url = req.session.selectedNode.settings.lnServerUrl + '/v1/newaddress?type=' + req.query.type;
-  request(options).then((body) => {
+  const qs: Record<string, any> = {};
+  if (req.query.type !== undefined) {
+    const raw = typeof req.query.type === 'string' ? req.query.type.trim() : '';
+    if (raw !== 'p2wkh' && raw !== 'np2wkh' && raw !== 'p2tr') {
```
saubyk, line 15: **Blocker: this breaks address generation for every LND user.**

The frontend never sends `p2wkh`/`np2wkh`/`p2tr`. `ADDRESS_TYPES` in `src/app/shared/services/consts-enums-functions.ts` carries `addressId: '0' | '1' | '4'` (the `lnrpc.AddressType` integers: `WITNESS_PUBKEY_HASH`, `NESTED_PUBKEY_HASH`, `TAPROOT_PUBKEY`), and `lnd.effects.ts:797` builds `NEW_ADDRESS_API + '?type=' + action.payload.addressId`. So after this change every "Generate Address" click on an LND node gets `400 type must be one of p2wkh, np2wkh, p2tr` and never reaches LND.

The accepted spellings are also wrong for the upstream: `p2wkh`/`np2wkh`/`p2tr` are `lncli newaddress` arguments, not values LND's REST gateway accepts for the `type` enum query parameter (it takes the enum name or its integer). The new `lnd-newAddress.test.mjs` cases for `type=p2tr`/`type=p2wkh` enshrine the wrong contract, and nothing exercises `type=1`.

The validation should accept what the UI sends (at minimum `0`, `1`, `4`, ideally the full `0`-`5` range plus the enum names), and the tests should use those values.

release-notes/Release-notes-0.15.12.md
```diff
@@ -170,6 +170,36 @@ this release should add its entry under the appropriate section below.
   still builds `?payment_hash=` by concatenation, as do `newAddress.ts`, `channels.ts` and
   `wallet.ts`.
 
+- **LND and Loop: the remaining query parameters were forwarded to the upstream as the
+  string `undefined`**
+  ([#TBD](https://github.com/Ride-The-Lightning/RTL/pull/TBD), fixes
```
saubyk, line 175: **Wrong release: this PR is based on `Release-0.15.12`, which shipped on 2026-09-08, and the note is in the 0.15.12 document.**

`CLAUDE.md` ("Branches and releases"): *"PRs target the current `Release-x.y.z` branch"* and *"Every PR adds its own release-note entry ... `release-notes/Release-notes-<x.y.z>.md`"*. The current branch is `Release-0.15.13` (60 commits ahead of `Release-0.15.12`), and `release-notes/Release-notes-0.15.13.md` already exists, so this entry belongs there under `## Bug Fixes`.

Please do **not** just retarget the PR in the GitHub UI. Release branches are rebase-merged, so `Release-0.15.13` shares no recent ancestor with `Release-0.15.12`; retargeting makes GitHub replay the entire 0.15.12 cycle into the PR and it goes `CONFLICTING` (see the "If a release ships while your PR is open" section of `CLAUDE.md`). Replay just this one commit instead:

```
git rebase --onto Release-0.15.13 Release-0.15.12 fix/1698-remaining-query-concat
```

then confirm `git diff Release-0.15.12..<old-head>` matches `git diff Release-0.15.13..<new-head>` before force-pushing, and retarget the PR to `Release-0.15.13`.

While you are there: the PR number is known now, so `#TBD` should become `#1709` (the convention is a follow-up `Fill in PR number in release note (#1709)` commit).

server/controllers/shared/loop.ts
```diff
@@ -64,9 +103,37 @@ export const loopOutTermsAndQuotes = (req, res, next) => {
   options.uri = '/v1/loop/out/terms';
   request(options).then((terms) => {
     logger.log({ selectedNode: req.session.selectedNode, level: 'DEBUG', fileName: 'Loop', msg: 'Loop Out Terms Received', data: terms });
-    const options1 = options; const options2 = options;
-    options1.uri = '/v1/loop/out/quote/' + terms.min_swap_amount + '?conf_target=' + (req.query.targetConf ? req.query.targetConf : '2') + '&swap_publication_deadline=' + req.query.swapPublicationDeadline;
-    options2.uri = '/v1/loop/out/quote/' + terms.max_swap_amount + '?conf_target=' + (req.query.targetConf ? req.query.targetConf : '2') + '&swap_publication_deadline=' + req.query.swapPublicationDeadline;
+    const qs: Record<string, any> = {};
```
saubyk, line 106: **Validation runs after the `/terms` round trip.** In both `*TermsAndQuotes` handlers the `targetConf`/`swapPublicationDeadline` checks sit inside `.then((terms) => ...)`, so a request with `targetConf=abc` still costs a full upstream call to the Loop server before the 400 is returned. The query is known before any I/O; parse it above `request(options)` (the same place `loopOutQuote`/`loopInQuote` do it) so a malformed request never touches the swap server. The existing test `loopOutTermsAndQuotes: targetConf=abc returns 400 without calling the quote endpoints` passes today only because it checks the *quote* endpoints, not that the terms call was skipped.

server/controllers/shared/loop.ts
```diff
@@ -46,10 +46,49 @@ export const loopOutTerms = (req, res, next) => {
...
+  if (amountRaw === '' || !(/^\d+$/).test(amountRaw)) {
+    logger.log({ selectedNode: req.session.selectedNode, level: 'WARN', fileName: 'Loop', msg: 'Invalid amount path param' });
+    return res.status(400).json({ message: 'amount must be a non-negative integer', error: 'Invalid path parameter' });
+  }
+  const amount = Number(amountRaw);
+  if (!Number.isSafeInteger(amount)) {
+    logger.log({ selectedNode: req.session.selectedNode, level: 'WARN', fileName: 'Loop', msg: 'amount exceeds safe integer range' });
+    return res.status(400).json({ message: 'amount exceeds maximum safe integer', error: 'Invalid path parameter' });
+  }
+  const qs: Record<string, any> = {};
```
saubyk, line 59: **The same ~30-line parse block is pasted four times in this file** (lines 59-91, 106-138, 201-233, 248-280 are byte-identical), and the inner "non-negative safe integer" shape is pasted again for `amount` (twice here), `target_conf` and `sat_per_vbyte` in `channels.ts`, `max_confs` in `wallet.ts`, on top of the copies #1687 left in `invoices.ts` and `payments.ts` (now ~12 in total). A single helper on `CommonService` in `server/utils/common.ts`, e.g. `parseNonNegativeInt(raw): number | null`, plus one `parseLoopQuoteQuery(req)` for the two loop fields, would collapse all of this and give the next handler with this shape somewhere to reuse. `eclair/invoices.ts:42` already has a private version of the integer parse that could be promoted.

test/backend/loop-quotes.test.mjs
```diff
@@ -0,0 +1,219 @@
...
+    await res.done;
+
+    assert.equal(res.statusCode, 200);
+    // 1 terms call + 2 quote calls
+    assert.equal(loop.seen.length, 3);
+    for (const req of loop.seen) {
+      assert.ok(!req.path.includes('undefined'), `URL contained undefined: ${req.path}`);
+    }
+    assert.equal(loop.seen[0].path, '/v1/loop/out/terms');
+    assert.equal(loop.seen[1].path, '/v1/loop/out/quote/100000?conf_target=2');
```
saubyk, line 199: **Order-dependent assertion on two concurrent requests.** `loopOutTermsAndQuotes` fires the min and max quote requests with `Promise.all`, on separate sockets; the fake server records them in arrival order, which is not guaranteed to match dispatch order. `seen[1]` = min / `seen[2]` = max will pass almost always and then fail once on a loaded CI box. Compare the two paths as a set instead, e.g. `assert.deepEqual(loop.seen.slice(1).map((s) => s.path).sort(), [...].sort())`.

test/backend/lnd-invoiceLookup.test.mjs
```diff
@@ -0,0 +1,127 @@
+import assert from 'node:assert/strict';
+import { createServer } from 'node:http';
+import test from 'node:test';
+
+import { invoiceLookup } from '../../backend/controllers/lnd/invoices.js';
```
saubyk, line 5: **Second test file for the same controller module, with the harness copied.** `test/backend/lnd-invoices.test.mjs` already imports from `backend/controllers/lnd/invoices.js` and defines the identical `startFakeLnd`/`buildRequest`/`buildResponse`; the same applies to the new `lnd-closeChannel.test.mjs` next to the existing `lnd-channels.test.mjs`. Adding the `invoiceLookup`/`closeChannel` cases to those files (the fake server there already takes the response body as a parameter) keeps one harness per controller instead of two, and matches how #1687 added `listInvoices`/`getPayments`.

SIDHARTH20K4:
@saubyk  fixed the blockers:

- `getNewAddress` now accepts the numeric IDs the UI sends (0, 1, 4), plus
  the enum names LND's REST gateway accepts. Empty and whitespace are still
  rejected. Tests updated to match.
- Release note moved to `release-notes/Release-notes-0.15.13.md` with #1709
  and removed from 0.15.12.
- Loop `TermsAndQuotes` handlers now parse the query before the `/terms`
  call, so a malformed `targetConf` or `swapPublicationDeadline` never
  reaches the swap server. Test asserts zero upstream calls on malformed
  input.
- Order-independent set comparison in `loop-quotes.test.mjs`.
- Promoted the integer parse from `eclair/invoices.ts` to
  `CommonService.parseNonNegativeInt` and reused it there. Same shape can
  replace the remaining copies in a follow-up if you want them in scope now.

Rebased onto `Release-0.15.13`, `git diff` before/after verified identical,
PR base retargeted to `Release-0.15.13`.

Issue #1710, opened 2026-09-21
Login fails the first time with "Invalid Csrf Token, Form Tempered"


bota87:
**Describe the bug**
When I open RTL and I type my password, the login fails with Invalid Csrf Token, Form Tempered.. If I type the same password again, it works.

It happens again every time I restart the container.

**To Reproduce**
1. Start a clean container:
   ```bash
   docker run -d --name rtl-test -p 3033:3000 shahanafarooqui/rtl:v0.15.12
   ```
2. Open `http://localhost:3033/rtl/` in a new browser profile.
3. Type the password and press Login.
4. The error appears. Type the same password again and the login works.

**Your environment**
* `RTL`: v0.15.12
* Browser: Brave

**Additional context**
In the browser console the login request returns HTTP 403.

saubyk:
Thanks for the clear report — reproduced, and the steps you gave are exactly right.

**What is happening**

The frontend takes its CSRF token from the `XSRF-TOKEN` cookie, and the server only sets that cookie in the catch-all handler that serves `index.html` for deep links (`/rtl/login`, `/rtl/home`, …). When you open `/rtl/` (or `/rtl`, which redirects there), the page is served by `express.static` instead, which sits above the catch-all and never mints a token. So the app loads with no token, the first login POST fails the CSRF check with 403, the server re-mints a token on that 403 response, and the second attempt succeeds.

That is also why it looks like a first-login-only problem: any refresh, bookmark or logout redirect lands on `/rtl/login`, which goes through the catch-all and works first time.

Verified against v0.15.12 with curl and in a clean browser profile:

| Entry URL | Token cookie set on load | First login |
|---|---|---|
| `/rtl/` | no | 403 |
| `/rtl/login` | yes | 200 |

**Why it shows up on every container restart**

Since 0.15.10 (`csurf` → `csrf-csrf`), tokens are signed with a secret generated at startup, so a token from a previous run is invalid after a restart. A fresh visit to `/rtl/` then doesn't mint a new one, and you hit the same 403 again. The static-page part of this is older than 0.15.12 (v0.15.9 fails the same way through curl), but the restart made it far more visible.

**Plan**

[...]

PR #1711, opened 2026-09-22
Mint the CSRF token for the landing page so the first login succeeds

saubyk:
Fixes #1710.

## Root cause

The frontend takes its CSRF token from the `XSRF-TOKEN` cookie, which `server/utils/app.ts` mints only in the catch-all that serves `index.html` for deep links (`/rtl/login`, `/rtl/home`, …). `GET /rtl/` — what people type, and where `/rtl` redirects — was answered by `express.static`, mounted above the catch-all, so the page loaded with no token and the first login POST failed the CSRF check with 403. The `EBADCSRFTOKEN` handler re-mints a token on that response, which is why the second attempt worked, and a refresh lands on `/rtl/login`, which is why a refresh worked first time.

Verified in a clean Chrome profile: all 13 requests the app makes on boot are served by `express.static` or the API; none reach the catch-all, so no token is ever set. The static-handler ordering was the same under `csurf` (v0.15.9 fails identically through curl), but since 0.15.10 tokens are signed with a per-boot secret, so every restart brought the 403 back — hence the reports now.

## Change

- `express.static` is mounted with `{ index: false }`, so the directory index falls through to the catch-all like every other page. `/rtl` still 301s to `/rtl/`; static files are served exactly as before.
- Not changed: a tab left open across an RTL restart still fails its first login once, since its token is void until the page reloads. Accepted, no automatic retry added.
- BTCPay SSO enters through `/rtl/api/authenticate/cookie`, which already fell through to the catch-all, so that path is unaffected. `docker/scripts/verify-sso.sh` needs no assertion changes; only its comment describing this trap is updated, as is the paragraph in `CLAUDE.md`.

## Test

`test/backend/csrf-landing-page.test.mjs` boots `rtl.js` in a child process with `NODE_ENV=production` (CSRF is disabled in development, and the static/catch-all order is fixed in the App constructor) and asserts:

- `GET /rtl/` returns `index.html` with `XSRF-TOKEN` and `_csrf` cookies
- a login after `GET /rtl/` succeeds on the first attempt
- a login without a token is still refused with 403
- `/rtl` still redirects to `/rtl/` and a static asset is still served with its own content type

Before the fix the first two fail (no cookie; 403). After: 4/4.

## Verification

- `npm run lint`: clean
- `npm run testbackend`: 163 passed, 0 failed
- Frontend Karma suite (`ng test`, ChromeHeadless): 230 passed, 0 failed
- `docker/scripts/verify-sso.sh` against the fixture with this build copied into `rtl-sso` and `rtl`: 16 passed, 0 failed (the stock v0.15.10 image fails only step 7, the #1656 malformed-body checks it predates). Also replayed the issue: `GET /rtl/` on the fixture RTL now sets both cookies and the first login returns 200; through the BTCPay nginx proxy `GET /rtl/` sets both cookies and `/rtl` still 301s to `/rtl/`.
- `backend/` regenerated with `npm run buildbackend`; `frontend/` untouched.


saubyk:
All four addressed in the follow-up commit:

- **F1** — the catch-all now sets `Cache-Control: no-store` before `sendFile`; the test asserts the header on `GET /rtl/`, and it is confirmed through the fixture's nginx proxy.
- **F2** — `boot()` rejects on the child's `'error'` event.
- **F3** — teardown awaits the child's exit (SIGTERM, SIGKILL after 5s) before removing the config directory.
- **F4** — added a control that a real `_csrf` cookie with an attacker-chosen `x-xsrf-token` is refused with 403.

Re-run: lint clean, `testbackend` 164/164, `verify-sso.sh` 16/16 with the new backend in the fixture.

saubyk:
Both addressed in the follow-up commit:

- **F5** — `GET /rtl/index.html` now 301s to `/rtl/` (handler mounted above `express.static`), so the token-minting catch-all is the single entry path. Test covers it; confirmed through the fixture's nginx proxy as well.
- **F6** — the boot timeout branch now sends SIGKILL.

Re-run: lint clean, `testbackend` 164/164, `verify-sso.sh` 16/16.

saubyk:
All five addressed in the follow-up commit:

- **F7** — both sides of the index-path comparison are case-folded (`baseHref` is a constant `/rtl` in `common.ts` today, so this is belt-and-braces).
- **F8** — the 301 sets `Cache-Control: no-store`; the test asserts it.
- **F9** — the `EBADCSRFTOKEN` 403 sets `Cache-Control: no-store`; the test asserts it on the no-token control.
- **F10** — comment corrected; confirmed against the fixture that `/rtl/%zz` is answered 200 by the catch-all.
- **F11** — `SKILL.md` now says 16.

Re-run: lint clean, `testbackend` 164/164, `verify-sso.sh` 16/16 with this backend in the fixture.

